PR #14, opened 2020-03-10
Add resource definiton for Okta Event Hooks

### LINKED ISSUE #20: okta_event_hook
*@quantumew commented on Sep 6, 2019, 5:36 PM UTC:*

Okta Event Hooks are now EA, add a resource for it.

[https://developer.okta.com/docs/reference/api/event-hooks/](https://developer.okta.com/docs/reference/api/event-hooks/)

*This issue was moved by [noinarisak](https://github.com/noinarisak) from [articulate/terraform-provider-okta#272](https://github.com/articulate/terraform-provider-okta/issues/272).*

mbudnek:
This adds a new `okta_event_hook` resource for managing Event Hooks.  The resource itself is mostly a copy/paste job from the existing `okta_inline_hook` resource.  Since the two are so similar I re-worked
"sdk/inline_hooks.go" to generalize it to work with both event hooks and inline hooks and renamed it to "sdk/hooks.go".

This should resolve #20.


mbudnek:
@noinarisak Any movement on this PR?  We're getting near the point we're going to start needing to use this functionality.

noinarisak:
Hi @mbudnek, Thanks for the PR! Sorry for the delay and I do agree this feature has been asked a few times. 🎉 

@quantumew and I, did a code review last week on this PR and had some concerns which are marked in this PR. Let us know if they make sense. @quantumew has been pretty busy too and hopefully, he can add some clearer examples that I mention in the review.

CHANGELOG.md
```diff
@@ -1,4 +1,9 @@
 ## 3.2.0 (Unreleased)
+
+FEATURES:
```
noinarisak: Please uncommit this file. Maintainers (myself) and Hashicorp peps update this file after post-CI/CD. Plus, I typically have to add more entries for other PR submitters that get merged. Thanks!

okta/resource_okta_event_hooks_test.go
```diff
@@ -0,0 +1,184 @@
...
+		unexpected := newSet.Difference(set)
+		if expected.Len() != 0 || unexpected.Len() != 0 {
+			return fmt.Errorf(
+				"%s: Attribute %s does not match expecation.  Missing values: %v, Unexpected values: %v",
+				resourceName,
+				attribute,
+				expected.List(),
+				unexpected.List(),
+			)
+		}
```
noinarisak: @quantumew and I thought this code was too complex. Actually, it was more of me, I'm slow. 😄  During the code review, @quantumew stated that he had a better example he would like to offer. Hopefully, @quantumew can paste that approach in this review.

sdk/inline_hooks.go
```diff
@@ -1,131 +0,0 @@
-package sdk
```
noinarisak: Both @quantumew and I had mix feeling about consolidating `Events` and `Inline` Hooks code. Typically, I lean toward DRY but because of the functionality between these two endpoints behavior different. `Events` being push base and `Inline` being trigger by process flow. I felt the concerns should be separated. Smaller code files vs large ones always make me more confident in changes with unit test, of course. 😄  

@quantumew also pointed out some sdk patterns that Okta sdk implemented during our code review but sadly I don't remember what that was. 😞 Hopefully, @quantumew can provide that feedback!

sdk/hooks.go
```diff
@@ -0,0 +1,211 @@
...
+		Channel *Channel `json:"channel"`
+		ID      string   `json:"id,omitempty"`
+		Name    string   `json:"name,omitempty"`
+		Status  string   `json:"status,omitempty"`
+		Type    string   `json:"type,omitempty"`
+		Version string   `json:"version,omitempty"`
+	}
+)
+
+func (e *Events) MarshalJSON() ([]byte, error) {
```
noinarisak: Could not find where this method was being used? Is it dead code?

mbudnek:
@noinarisak I don't see any review comments on this PR.  Did you forget to submit the review maybe?  Or I think GitHub had an outage earlier, maybe that ate the review.

noinarisak (changes requested):
Hi @mbudnek. User error! 😃  GitHub review was waiting for me hit the `Submit Review` button. 

The following our feedback. Hopefully, @quantumew can add his feedback and examples!

okta/resource_okta_event_hooks_test.go
mbudnek: I agree, this is complicated.  If there's a simpler way to do this I'd love to use it, but this is the best I could come up with given my limited experience with both Go and the Terraform plugin SDK.  I did came across `testCheckResourceSliceAttrForSet`, and that works in most cases.  It doesn't work for sets of maps though as is the case for the "headers" attribute.

sdk/inline_hooks.go
mbudnek: I definitely _can_ split these back apart, but I'm not thrilled with the idea.  It will mean duplicating every type and function in this file with the only difference being replacing the word 'inline' with 'event'.  Even worse, the `AuthScheme`, `Channel`, `HookConfig`, and `Header` types will be completely identical.  That means either re-naming the existing ones (to something like `InlineHookAuthScheme`) and touching the inline hook resource (which I was trying to avoid) or leaving them named inconsistently.

Even beyond that, I'm not sure I agree that inline and event hooks are really fundamentally different things.  They're both just callbacks and only really differ in whether they're called synchronously or asynchronously (and the types of events you can get called back for).

Again, I definitely will do so, if you and @quantumew think it's the right way to go.  Just wanted to give my reasoning here.

sdk/hooks.go
mbudnek: This makes the `Events` type to implement the [`Marshaler`](https://golang.org/pkg/encoding/json/#Marshaler) interface so that `json.Marshal` calls this function when marshaling an `Events` object.  It's here just to inject static value `"EVENT_TYPE"` as the `type` key in the marshaled JSON object since that key should never have any other value (currently, at least).

Looking at it now, it may be overly-clever.  I can add an explicit `Type` member to the `Events` class and get rid of this if you would prefer.

okta/resource_okta_event_hooks_test.go
mbudnek: I simplified this a little bit and added some comments to explain what it's doing.

mbudnek:
@noinarisak @quantumew I split hooks.go back out into inline_hooks.go and event_hooks.go and cleaned up some of the stuff from @noinarisak's review.  Let me know what you think.

mbudnek:
@noinarisak @quantumew Any further movement on this?

noinarisak (approved):
@mbudnek Looks great!

noinarisak:
@mbudnek Sorry for being MIA. It's been crazy the last couple of months. This PR looks solid. Thanks again for the changes!

Issue #93, opened 2020-03-25
Optimize app retrieval by ID 


andymarch:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

data_source_okta_app uses the /apps end point with a filter to identify apps regardless of the information provided in the filter. The rate limit for create and list applications is significantly lower than most other endpoints (25 for most orgs, going up to 100 with enterprise license). There is a separate rate limit for /apps/{id} which falls into a different and much larger rate limit bucket. 

This is specifically called out as the example on the[ rate limits page](https://developer.okta.com/docs/reference/rate-limits/).

> For example, the limit for getting an application by ID is higher than the more general limit for the /api/v1/apps endpoint.

If the provided filter contains an id then the more specific /apps/{id} should be used rather than using the filtered list.

Happy to take a run at this myself, it looks like the best place for a change would be in [app_filter](https://github.com/terraform-providers/terraform-provider-okta/blob/9ae5cc400dab482fc4d43a798e1295b00712185c/okta/app_filter.go).

### New or Affected Resource(s)

* okta_app

### Potential Terraform Configuration

```hcl
# Should list all applications with a filter on the label
data "okta_app" "anapp" {
  label = "An App"
}

# Currently this performs a list with an id filter
data "okta_app" "anotherapp" {
  id = "0oa9k05yf3gGKOEV30x6"
}
```

### References

* #35 - Indicates similar issues with managing app users.

noinarisak:
@andymarch Could you let us know if PR #97 resolved this issue. Thanks!

adw1n:
@noinarisak it did not. I'm using the latest v3.4.1 relase and I see a ton of requests being made to /api/v1/apps which consitently causes rate limit violations for my Org. This furthermore causes downtime for my app which relies on being able to use `/api/v1/apps` endpoint. I had a look at https://github.com/oktadeveloper/terraform-provider-okta/pull/97 description and

> Improve app filtering by using ?q=<label> rather than listing and filtering. App doesn't have good filtering like users do so this was the best approach.

apporach sounds really problemtatic. The way I see it there at least three approaches:
[...]

Issue #95, opened 2020-03-26
Infinite loop on okta_app data source


PurrBiscuit:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

```
terraform -v
Terraform v0.12.23
+ provider.aws v2.54.0
+ provider.consul v2.6.1
+ provider.datadog v2.7.0
+ provider.okta v3.0.37
+ provider.vault v2.9.0
```

### Affected Resource(s)

* [data.okta_app](https://www.terraform.io/docs/providers/okta/d/app.html)

### Terraform Configuration Files

```hcl
data "okta_app" "scim_import" {
  label = "Example OIN Import"
}
```

### Debug Output

```
2020-03-26T15:19:07.477Z [DEBUG] plugin.terraform-provider-okta_v3.0.37_x4: GET /api/v1/apps??after=0oancj172zSy5Y2hJ0h7&filter=status+eq+%22ACTIVE%22&limit=200 HTTP/1.1
2020-03-26T15:19:11.235Z [DEBUG] plugin.terraform-provider-okta_v3.0.37_x4: GET /api/v1/apps??after=0oancj172zSy5Y2hJ0h7&filter=status+eq+%22ACTIVE%22&limit=200 HTTP/1.1
2020-03-26T15:19:15.693Z [DEBUG] plugin.terraform-provider-okta_v3.0.37_x4: GET /api/v1/apps??after=0oancj172zSy5Y2hJ0h7&filter=status+eq+%22ACTIVE%22&limit=200 HTTP/1.1
2020-03-26T15:19:20.046Z [DEBUG] plugin.terraform-provider-okta_v3.0.37_x4: GET /api/v1/apps??after=0oancj172zSy5Y2hJ0h7&filter=status+eq+%22ACTIVE%22&limit=200 HTTP/1.1
```

### Expected Behavior

The data source should find the app by `label` name and return the resulting `id`.

### Actual Behavior

The `terraform plan` run gets stuck waiting for the `okta_app` data source to error or return a result.  Eventually have to kill the process altogether.

[...]

PR #97, opened 2020-03-26
Improve app filtering and update Terraform SDK

quantumew:
* Update Terraform SDK to 1.9.0
* Improve app filtering by using `?q=<label>` rather than listing and filtering. App doesn't have good filtering like users do so this was the best approach. 
* Avoid endless loops when our after param is always the same

Related to #95, doesn't quite fix it. There is a little more to do here, just getting a bug fix in place.


noinarisak:
Updating `?q=<label>` approach should also resolve #93.

okta/app_filter.go
```diff
@@ -41,13 +41,14 @@ func (a *appFilters) String() string {
 
 func listApps(m interface{}, filters *appFilters) ([]*appID, error) {
 	result := &searchResults{Apps: []*appID{}}
-	qp := &query.Params{Limit: 200, Filter: filters.ApiFilter}
+	qp := &query.Params{Limit: 200, Filter: filters.ApiFilter, Q: filters.getQ()}
```
quantumew, line 44: Now that we are using `?q=` for the label, we should document the fact that this will gather results that match via name and label.

quantumew: 3e1e5eeb439f6ac1e28c70eaefbe9b8f5aecf75e

noinarisak:
@quantumew Could you also rebase your commits to single commit. Just following the standard that I'm telling every PR submitter to keep our git history tidy. =) 

Thanks!

quantumew:
@noinarisak Do you not have the option to do a rebase or squash merge? 

https://github.blog/2016-09-26-rebase-and-merge-pull-requests/

noinarisak:
Thanks for the rebase, @quantumew ! I have access to all 3 options but maybe bc of my past experience (or being a novice) it was a pain when I tried to revert a PR with rebase and/or squash merge. Not sure if you had similar issues

PR #106, opened 2020-04-07
Remove default value for key_years_valid in okta_app_saml

SebastianCzoch:
I've tested in on production, nothing change for current configuration (where I had to set this value because default value was 1 which is incorrect), also working perfectly fine for config where I removed this field.

IMO there is no need to test it in unit tests because we are using external library (helpers from terraform plugins sdk which should be tested enough).

This issue closing #73 

This is not a braking change

### DIFF
--- a/okta/resource_okta_app_saml.go
+++ b/okta/resource_okta_app_saml.go
@@ -68,7 +68,6 @@ func resourceAppSaml() *schema.Resource {
 			"key_years_valid": {
 				Type:         schema.TypeInt,
 				Optional:     true,
-				Default:      1,
 				ValidateFunc: validation.IntAtLeast(1),
 				Description:  "Number of years the certificate is valid.",
 			},


SebastianCzoch:
@noinarisak Is there any chance to get it reviewed and merged?

merlin-zdn (approved):
LGTM

netflash:
Is there any chance to get it reviewed and merged?

netflash:
Can we please merge this?

SebastianCzoch:
@netflash probably this repository is dead...
@noinarisak is that true?

noinarisak:
Hi @SebastianCzoch project is still going, I personally have been really busy since I'm only a volunteer for this project. The good news, I'm getting some fulltime dedicated help going forward. The individual is still onboarding, etc. 😄  

In the meantime, I got some cycles next week and review the associated Issue but sounds like there is not to much harm with our PR. However, I believe @quantumew was concern this would impact existing SAML apps mention in #73.

SebastianCzoch:
Hey @noinarisak thanks for the reply!

As I mentioned in PR description, I was testing various cases here and everywhere it works as expected (also when migrating configuration from okta provider without this change to this)

bogdanprodan-okta:
Hi @SebastianCzoch ! The problem is that when `key_years_valid` doesn't have a default value and there is `key_name` property specified in the terraform file, the execution will error out. The reason for this is that in the official documentation for [generating key credential](https://developer.okta.com/docs/reference/api/apps/#generate-new-application-key-credential) number of years is a required parameter and not specifying it will cause an error: `Error: The API returned an error: Api validation failed: generateKey. Causes: errorSummary: Validity years out of range. It should be 2 - 10 years`
Yes, the default value is incorrect and should be `2` as well as `ValidateFunc` should be `validation.IntBetween(2, 10)`
The fix for this issue is included in the #215 
If you have more questions or concerns, I'm always open for discussion.

bryantbiggs:
@noinarisak  / @bogdanprodan-okta I think we can close this now with #215 ?

bogdanprodan-okta:
Hi @SebastianCzoch! I'm closing this PR as a bug regarding `key_years_valid` was fixed in #215. If you have some other concerns, please write your comment and we always can reopen this PR. Thanks!

PR #120, opened 2020-05-08
Expose Implicit Assignment [Federation Broker] setting of okta_app_oauth resource

jlew:
This enables the ability to turn on [Federation Broker Mode](https://help.okta.com/en/prod/Content/Topics/Apps/Apps_Apps_Page.htm#Enable) for an application.

> Federation Broker Mode allows for SSO without the need to pre-assign apps to specific users. Access is managed only by sign-on policy and the authorization rules of each app. This mode can improve import performance and can be especially helpful for larger-scale orgs that manage many users and/or apps.

# Example Usage
```terraform
resource okta_app_oauth app {
  label          = "123"
  status         = "ACTIVE"
  redirect_uris  =  "https://localhost:3000/"

  type           = "browser"
  grant_types    = ["authorization_code"]
  response_types = [ "code" ]
  token_endpoint_auth_method =  "none" 
  consent_method = "TRUSTED"

  // Turn on Federation Broker Mode
  implicit_assignment = true
}
```

# Notes about usage
* Note that this feature *"is an Early Access feature"* and needs to be enabled by Okta Support before this flag can be used.
* When enabled, the `users` and `groups` arguments will stop being honored for this resource as the okta backend will fail out those requests.
  * Other resources that create assignments that are not in line (i.e. `okta_app_group_assignments`) will also fail:
     > Error: The API returned an error: Api validation failed: GroupAppAssignment. Causes: errorSummary: It is not possible to assign groups to an AppInstance that has Federation Broker Mode enabled

# Note About Test
I have provided a test to ensure this code works, however since this is behind a feature flag on Okta's side I have configured it to skip.

Current state:
```
make testacc TEST=./okta TESTARGS='-run=TestAccAppOauth_federationBroker'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./okta -v -run=TestAccAppOauth_federationBroker  -timeout 120m
=== RUN   TestAccAppOauth_federationBroker
--- SKIP: TestAccAppOauth_federationBroker (0.00s)
    resource_okta_app_oauth_test.go:118: This is an 'Early Access Feature' and needs to be enabled by Okta, skipping this test as it fails when this feature is not available
PASS
ok  	github.com/terraform-providers/terraform-provider-okta/okta	(cached)
```

When run against passing account (without the skip):
```
make testacc TEST=./okta TESTARGS='-run=TestAccAppOauth_federationBroker'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./okta -v -run=TestAccAppOauth_federationBroker  -timeout 120m
=== RUN   TestAccAppOauth_federationBroker
--- PASS: TestAccAppOauth_federationBroker (11.21s)
PASS
ok  	github.com/terraform-providers/terraform-provider-okta/okta	11.693s
```

When Run against account without feature (without the skip):
```
make testacc TEST=./okta TESTARGS='-run=TestAccAppOauth_federationBroker'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./okta -v -run=TestAccAppOauth_federationBroker  -timeout 120m
=== RUN   TestAccAppOauth_federationBroker
--- FAIL: TestAccAppOauth_federationBroker (10.95s)
    testing.go:669: Step 1 error: Check failed: Check 5/5 error: okta_app_oauth.test: Attribute 'implicit_assignment' expected "true", got "false"
FAIL
FAIL	github.com/terraform-providers/terraform-provider-okta/okta	12.718s
make: *** [testacc] Error 1
```


bogdanprodan-okta:
Hi @jlew! Thanks for the PR and for the late response. I need to make an investigation of whether Okta supports this feature because based on the [documentation](https://developer.okta.com/docs/reference/api/apps/), I can't find any mention of `implicitAssignment` field.

jlew:
Looks like Okta has changed their documentation link since the pr was open.

A quick search:
[Federation Broker Mode](https://help.okta.com/en/prod/Content/Topics/Apps/apps-fbm-main.htm)

bogdanprodan-okta:
I'll try to figure out whether this is still supported so we can either close or merge this PR.

jlew:
We currently use this in our enterprise deployment and therefore we have not been using the mainline provider.  Instead we have been maintaining a patched version as our usage of Okta does not perform without this configuration.  Okta doesn't scale well when you are using "social providers" where all users are federated with another system.  The just in time user creation system gets backed up adding the users to each application.  When this mode is enabled it changes how Okta associates users to the application.  For most use cases where users are managed in Okta, this not a problem.  When Okta is used as an authorization server and not an identity server at scale, things get slow and users cannot use the app until Okta gets caught up provisioning federated users and as the number of apps increase, adding one user to each app that is assigned the everybody group now causes them to be unable to login if this is the first time they federated in for quite some time.

Okta does document the existence of the feature but not in the API documents as you need to reach out to your account manager or support to enable the feature.  You can also see this check box on the UI for each application once this feature is enabled.

This is why the pr calls out specifically that you need to get the feature enabled on the test account for the tests to pass.

So we would need the accounts that the tests run against have this option enabled or disable that specific test.

If we are not able to merge this currently, I would prefer to see the pr remain open in a blocked state waiting for Okta to general release the Federation Broker Mode instead simply close it so that it is visible to other enterprises that may run into this and can be revived if/when Okta makes this an official feature.

Next week I can reach out to our Okta representative and see if we can find an official document about this feature.

nimeshsubramanianv2:
@bogdanprodan-okta @jlew Can we get this PR merged?

bogdanprodan-okta:
Hi everyone! This feature will be available in the upcoming v3.8.0. Stay tuned!

PR #126, opened 2020-05-22
Fixed provider_id typo in okta_factor in web documentation

ChristophShyper:
Taken from https://github.com/articulate/terraform-provider-okta/blob/master/okta/resource_okta_factor.go and https://github.com/articulate/oktasdk-go/blob/master/okta/org.go

Web documentation is misleading and wrong. `provider` is a reserved keyword and will create errors in Terraform that will be connected with the Okta provider and not this particular resource. Hence will be very confusing for beginners.

Checking in code its name is `provider_id` and have a list of allowed values.

I'm fixing those two important information here.

### DIFF
--- a/website/docs/r/factor.html.markdown
+++ b/website/docs/r/factor.html.markdown
@@ -16,7 +16,7 @@ This resource allows you to manage Okta MFA methods.
 
 ```hcl
 resource "okta_factor" "example" {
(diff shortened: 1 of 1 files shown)


noinarisak:
@ChristophShyper Thanks!

PR #132, opened 2020-06-03
Added missing argument description

lasektom:
Adding the missing description of `grant_type_whitelist` argument. Spent some time researching how it works especially that I needed to use it in conjunction with the EVERYONE group.

### DIFF
--- a/website/docs/r/auth_server_policy_rule.html.markdown
+++ b/website/docs/r/auth_server_policy_rule.html.markdown
@@ -40,6 +40,8 @@ The following arguments are supported:
 
 * `priority` - (Required) Priority of the auth server policy rule.
 
+* `group_whitelist` - (Required) The list of group ids to whitelist. Can be set to group id or to the following: "EVERYONE".
+
 * `grant_type_whitelist` - (Required) Accepted grant type values, `"authorization_code"`, `"implicit"`, `"password"`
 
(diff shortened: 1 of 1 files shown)


website/docs/r/auth_server_policy_rule.html.markdown
```diff
@@ -40,6 +40,8 @@ The following arguments are supported:
 
 * `priority` - (Required) Priority of the auth server policy rule.
 
+* `group_whitelist` - (Required) The list of group ids to whitelist. Can be set to group id or to the following: "EVERYONE".
```
thekbb, line 43: "EVERYONE" should be styled with inline code:  `"EVERYONE"`

noinarisak: @lasektom I agree with @thekbb. Can you update the markdown like this?

```markdown
`group_whitelist` - (Required) The list of group ids to whitelist. Can be set to group id or to the following: `"EVERYONE"`.
```

noinarisak (changes requested):
@lasektom Could you update your commit with our request. Thanks!

bryantbiggs:
@noinarisak / @bogdanprodan-okta I think this can be closed since the docs already have the necessary updates 
image

PR #135, opened 2020-06-11
 commiting changes to support exsiting authorization server claim and…

patluri43:
Added support for Existing Claims and policies

### DIFF
--- a/examples/okta_network_zone/basic_updated.tf
+++ b/examples/okta_network_zone/basic_updated.tf
@@ -1,6 +1,7 @@
 resource "okta_network_zone" "ip_network_zone_example" {
   name     = "testAcc_replace_with_uuid Updated"
   type     = "IP"
+  blacklist = false
   gateways = ["1.2.3.4/24", "2.3.4.5-2.3.4.10"]
   proxies  = ["2.2.3.4/24", "3.3.4.5-3.3.4.10"]
 }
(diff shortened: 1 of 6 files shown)


bogdanprodan-okta:
Hi @patluri43! Thanks for the PR and sorry for the late response. I've made an investigation, and it's probably better to just import default claim or policy, rather than do it this way. Basically, you just create an empty resource in the terrafrom file:
```tf
resource "okta_auth_server_claim" "test" {

}
```
and then execute `terraform import okta_auth_server_claim.test {auth_server_id}/{claim_id}`.
Statefile:
```json
{
      "mode": "managed",
      "type": "okta_auth_server_claim",
      "name": "test",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "always_include_in_token": true,
            "auth_server_id": "auspqfzqjoRAoHNZP0k7",
            "claim_type": "IDENTITY",
            "group_filter_type": "",
            "id": "odlpqfzqk1Ljp0BOi0h7",
            "name": "email",
            "scopes": [
              "email"
            ],
            "status": "ACTIVE",
            "value": "",
            "value_type": "SYSTEM"
          },
          "private": "eyJzY2hlbSFfdmVyc2lvaiI6IjDifQ=="
        }
      ]
    },
```
The only problem that ` "value_type": "SYSTEM"` is not supported by the provider, so I think I'll add the functionality to support system claims import, but disable create, delete and update (the only field, that can be changed for the system claims is `always_include_in_token` with the only exception: `sub` claim, which can not be changed at all)

PR #137, opened 2020-06-23
add support for Federated Broker mode (implicit assignment) and acs_binding fix

eatplaysleep:
Okta provides what they call a [Federated Broker Mode](https://help.okta.com/en/prod/Content/Topics/Apps/Apps_Apps_Page.htm#) that is currently an Early Access feature but supported in the SDK. 

PR adds support for this flag as well as fixes an issue where the acs_binding value was getting set even though it is not a valid value on an OIDC/OAuth app in Okta. 

Sets it to optional so that if it is not present in the configuration it simply does not get set.


eatplaysleep:
Found a minor bug in the auth server police rule import/refresh so committed a quick fix.

bogdanprodan-okta:
Hi @eatplaysleep! Thanks for your PR! Seems to me, that this PR is a duplicate of #120. In addition, it's an `Early Access` feature, so we can only add this feature after Okta officially releases it.
 It's also very complicated to review, as it's huge and contains a lot of unrelated changes. If you want to fix some bugs or add some extra functionality, please create another PR.

eatplaysleep:
> Hi @eatplaysleep! Thanks for your PR! Seems to me, that this PR is a duplicate of #120. In addition, it's an `Early Access` feature, so we can only add this feature after Okta officially releases it.
> 
>  It's also very complicated to review, as it's huge and contains a lot of unrelated changes. If you want to fix some bugs or add some extra functionality, please create another PR. 

@bogdanprodan-okta makes sense! I'll follow #120. This is both an old one and messy one and I need to go back and revisit everything I did. 

I'll submit something new if applicable still. 

Thank you!

PR #143, opened 2020-07-01
Add group_whitelist to schema & make scope_whitelist required

ymylei:
This PR is for a number of related items:

1. Adds "group_whitelist" into the schema and the docs, it's already used in the examples however not documented and fully implemented in the resource scheme. All of the condition object items are worth a re-visit to cleanup, ideally after Okta SDK upgrade, since there's odd behavior like using the "GroupRule" conditionals versus the generic conditionals. We can also add "user_whitelist" and "{type}_blacklist" as well.

2. This makes scope_whitelist required in the resource scheme to match the docs. Technically the API does not require it, however since the default for this is empty, the created resource is unusable in that state (according to my testing).

3. I've updated the policy rule examples / tests to include scope_whitelist (also add auth code to the updated test since that'll be one of the most used settings for these).

4. I've updated the docs around "grant_type_whitelist" to reflect that "client_credentials" is also valid.

5. I ran `terraform fmt` on the examples DIR while working on this.

### DIFF
--- a/examples/okta_app_oauth/service_with_jwks.tf
+++ b/examples/okta_app_oauth/service_with_jwks.tf
@@ -1,8 +1,8 @@
 resource "okta_app_oauth" "test" {
-  label          = "testAcc_replace_with_uuid"
-  type           = "service"
-  response_types = ["token"]
-  grant_types    = ["client_credentials"]
+  label                      = "testAcc_replace_with_uuid"
+  type                       = "service"
+  response_types             = ["token"]
+  grant_types                = ["client_credentials"]
   token_endpoint_auth_method = "private_key_jwt"
 
   jwks {
--- a/examples/okta_auth_server_policy_rule/basic.tf
+++ b/examples/okta_auth_server_policy_rule/basic.tf
@@ -10,6 +10,7 @@ resource "okta_auth_server_policy_rule" "test" {
   priority             = 1
   group_whitelist      = ["${data.okta_group.all.id}"]
   grant_type_whitelist = ["implicit"]
+  scope_whitelist      = ["openid","profile"]
 }
 
 resource "okta_auth_server" "test" {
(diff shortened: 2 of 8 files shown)


ymylei:
This also contains a dupe of the PR #132

ymylei:
Acceptance Test Output:
```
tom@DESKTOP-NNSTVLT:~/terraform-provider-okta$ make testacc TEST=./okta TESTARGS='-run=TestAccOktaAuthServerPolicyRule_create'==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./okta -v -run=TestAccOktaAuthServerPolicyRule_create  -timeout 120m
=== RUN   TestAccOktaAuthServerPolicyRule_create
--- PASS: TestAccOktaAuthServerPolicyRule_create (7.88s)
PASS
ok      github.com/terraform-providers/terraform-provider-okta/okta     (cached)
```

bogdanprodan-okta:
Hi @ymylei ! Thanks for submitting this PR! I think you've missed the fact that schema for this resource is created by calling
`addPeopleAssignments` [function here](https://github.com/ymylei/terraform-provider-okta/blob/auth-server-pol-rule-docs/okta/resource_okta_auth_server_policy_rule.go#L20). It adds `peopleSchema` map and the resulting schema has all the necessary parameters.
Here is the list of current schema keys: 
1. grant_type_whitelist
2. policy_id
3. group_blacklist
4. access_token_lifetime_minutes
5. refresh_token_lifetime_minutes
6. scope_whitelist
7. user_whitelist
8. auth_server_id
9. priority
10. type
11. refresh_token_window_minutes
12. user_blacklist
13. group_whitelist
14. name
15. nline_hook_id
16. status

However, you are right about missing docs, I'll add them in #215

PR #147, opened 2020-07-10
Fix validation false positive when api_token is set via environment variable

jgeurts:
Terraform validate fails when api_token is not specified but `OKTA_API_TOKEN` environment variable is specified. This should fix validation in that case. If this is not the preferred method to fix that bug, please let me know and I'd be happy to alter the change accordingly!


noinarisak:
Thanks @jgeurts! After some conversation with my teammate, it makes more sense for it to be optional and also the `org_name`. Updating docs and more.

noinarisak:
Related to PR #208

PR #151, opened 2020-07-16
Update app_oauth.html.markdown

noinarisak:
Update documentation for `"private_key_jwt:` in `resource_app_oauth` becuase it was missing.

### DIFF
--- a/website/docs/r/app_oauth.html.markdown
+++ b/website/docs/r/app_oauth.html.markdown
@@ -24,6 +24,23 @@ resource "okta_app_oauth" "example" {
 }
 ```
 
+```hcl
+resource "okta_app_oauth" "example" {
+  label          = "example"
+  type           = "service"
+  response_types = ["token"]
+  grant_types    = ["client_credentials"]
+  token_endpoint_auth_method = "private_key_jwt"
+
+  jwks {
+    kty = "RSA"
+    kid = "SIGNING_KEY"
+    e   = "AQAB"
+    n   = "xyz"
+  }
+}
+```
+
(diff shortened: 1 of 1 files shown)


website/docs/r/app_oauth.html.markdown
```diff
@@ -44,7 +44,7 @@ The following arguments are supported:
 
 * `client_basic_secret` - (Optional) OAuth client secret key, this can be set when token_endpoint_auth_method is client_secret_basic.
 
-* `token_endpoint_auth_method` - (Optional) Requested authentication method for the token endpoint. It can be set to `"none"`, `"client_secret_post"`, `"client_secret_basic"`, `"client_secret_jwt"`.
+* `token_endpoint_auth_method` - (Optional) Requested authentication method for the token endpoint. It can be set to `"none"`, `"client_secret_post"`, `"client_secret_basic"`, `"client_secret_jwt"`, `"private_key_jwt"`.
```
varrunramani-okta, line 64: It would be nice for the documentation also include an example of how to use `jwks` as shown in [this file](https://github.com/terraform-providers/terraform-provider-okta/blob/master/examples/okta_app_oauth/service_with_jwks.tf)

noinarisak: @varrunramani-okta I could add more usage examples in the docs at the top of the page

varrunramani-okta: I think that would be helpful and/or adding `jwks` to the argument/attribute reference (not sure what it falls under)

PR #153, opened 2020-07-19
[WIP] Upgrade Provider to Okta SDK v2

ymylei:
WIP!
[SDK v2 Docs](https://pkg.go.dev/github.com/okta/okta-sdk-golang/v2/okta?tab=doc)

This PR is designed to update the provider to use the latest v2 version of the Okta Golang SDK. So far I've:
1. Converted all usages of v1 SDK to v2
2. Filled in all the calls that want context in the new version using the context generated on client load
3. I've had to change the config of the client/provider itself to use the correct timeout and retry parameters (BREAKING CHANGE)
4. I've updated the native functions used where the name or types of object used in the api calls changes (None appear to be breaking so far)
5. Modified the supplemental SDK to also use the new native Do functions. I've marked Admin roles and Idps as ones I will remove and use the native SDK for with this PR.
6. Some linting for the non variable name stuff.
7. Migrate admin roles API calls to native SDK
8. Time thing no longer a big deal, it converts to time.Time in the SDK itself so we should document to use ints as seconds and let the lib handle it.

TODO:
1. SAML attribute statements are messed up currently (we only have a definition for one type, however the API itself recognizes "regular" attributes and "group" attributes separately, I will try to deal with that here) <- Likely to also be a breaking change
2. Convert IDP supplement calls to native SDK
3. I will also try and completely remove all use of the Articulate Okta SDK from this, since we should probably let that library die for better or for worse.
4. Will update tests, docs, and associated items for the breaking change items as well as any minor changes which need updating.
5. Other lints items that aren't var name casing

Upstream PR deps:
1. [int64 for timeouts](https://github.com/okta/okta-sdk-golang/pull/151)
2. [cache lib use fix](https://github.com/okta/okta-sdk-golang/pull/150)
3. [policy rule obj name](https://github.com/okta/okta-sdk-golang/pull/152)


ymylei:
Moved the admin api calls over, also linked out to a couple of the sdk fix PRs I we'll need integrated in here as well.

noinarisak:
Hey @ymylei ! Just a heads up. We got a new member to the team, that will be fulltime vs me (volunteer/parttime) on the project and is taking over the migration to v2. He is dedicated to implementing the migration and hopefully gets something in the next couple of weeks.

I'll keep this PR open in case there is some reference information, but will most likely close it after we merge his changes in.

bryantbiggs:
I think this PR can be closed since this https://github.com/oktadeveloper/terraform-provider-okta/pull/203 was just merged

PR #155, opened 2020-07-20
Adding type to data okta group

dangoslen:
This PR adds the ability to filter by a `type` in the `okta_group` data entity. 

This is useful in cases were Okta is managed groups from multiple sources (LDAP, Google groups, etc.) that have the same name but different types. 

It additionally returns the type as an attribute of the entity.


tgoodsell-tempus:
@noinarisak Any chance we could get this an expedited review? I've found that being able to set the `type` parameter on group api queries is a lot more important once you start having a ton of group push provisioning being done (since they'll create groups of the same name based on what's in your application).

okta/data_source_okta_group.go
```diff
@@ -18,6 +18,12 @@ func dataSourceGroup() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 			},
+			"type": &schema.Schema{
+				Type:        schema.TypeString,
+				Optional:    true,
+				Default:     false,
+				Description: "Type of the group. When specified in the terraform resource, will act as a filter when searching for the group",
+			},
```
bogdanprodan-okta, line 27: Could you please add a validation function for group types?
https://developer.okta.com/docs/reference/api/groups/#group-type

okta/data_source_okta_group.go
```diff
@@ -44,7 +50,13 @@ func dataSourceGroupRead(d *schema.ResourceData, m interface{}) error {
 
 func findGroup(name string, d *schema.ResourceData, m interface{}) error {
 	client := getOktaClientFromMetadata(m)
-	groups, _, err := client.Group.ListGroups(&query.Params{Q: name})
+	searchParams := &query.Params{Q: name}
+	if d.Get("type") != nil {
+		groupType := d.Get("type").(string)
+		searchParams = &query.Params{Q: name, Type: groupType}
+	}
```
bogdanprodan-okta, line 57: This can be simplified
```	searchParams := &query.Params{Q: name}
	if d.Get("type") != nil {
		searchParams.Type = d.Get("type").(string)
	}
```

dangoslen:
Sorry - I royally screwed up the merge from master down. Will open a new PR if that is ok with you @bogdanprodan-okta

bogdanprodan-okta:
Sure, go ahead!

PR #156, opened 2020-07-28
Adds name to the error for group data call

ymylei:
When this is embedded inside of modules, the error message becomes very difficult to use since the actual call is abstracted. By adding the name of the group to the error message, it makes it much easier to discover which you're not referencing correctly.


ymylei:
If necessary, I'm willing to wait for #155 to be merged in and I'll rework these logging changes based upon those changes. I think that should take priority over this in terms of merge ordering.

tgoodsell-tempus:
@bogdanprodan-okta @noinarisak Would you be able to include this in the next patch release?

tgoodsell-tempus:
@bogdanprodan-okta much appreciated!

PR #158, opened 2020-07-29
Handling for INVALID Group Rule State

ymylei:
This is providing some handling when the group rule is detected as invalid (when the group it's meant to assign you to has been deleted).


ymylei:
Designed to resolve  #158

PR #165, opened 2020-08-11
Update app_group_assignment.html.markdown

snolan-amount:
Corrected the help for importing, as per the help in the CLI, you need the app_id and group_id not the assignment id.


noinarisak:
LGTM! Thanks, @snolan-amount !

PR #172, opened 2020-08-20
Update warning in app_group_assignment.html.markdown

ssttgg:
This caught me out and I found the documentation slightly ambiguous  so I thought I would really spell out the fix and the reason why it might be required.

### DIFF
--- a/website/docs/r/app_group_assignment.html.markdown
+++ b/website/docs/r/app_group_assignment.html.markdown
@@ -26,7 +26,7 @@ JSON
 }
 
 ```
-__NOTE__: When using this resource in conjunction with other application resources (e.g. `okta_app_oauth`) it is advisable to add the following `lifefycle` argument to the associated `app_*` resources to prevent the groups being unassigned on subsequent runs:
+__NOTE__: When using this resource in conjunction with other application resources (e.g. `okta_app_oauth`) it is advisable to add the following `lifecycle` argument to the associated `app_*` resources to prevent the groups being unassigned on subsequent runs:
 
 ```hcl
 resource "okta_app_oauth" "app" {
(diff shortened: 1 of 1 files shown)


website/docs/r/app_group_assignment.html.markdown
```diff
@@ -32,6 +24,17 @@ resource "okta_app_group_assignment" "example" {
 }
 JSON
 }
+
+```
+__NOTE__: When using this resource in conjunction with other application resources (e.g. `okta_app_oauth`) it is advisable to add the following `lifecycle` argument to the associated `app_*` resources to prevent the groups being unassigned on subsequent runs:
```
noinarisak: Thanks, @ssttgg for the PR! 

I agree with PR. However, could you re-format the markdown like the example below? Just following HashiCorp markdown [callout standards](https://www.terraform.io/docs/registry/providers/docs.html#calloutsl). 😄 

```markdown
!> **NOTE** When using this resource in conjunction with other application resources (e.g. `okta_app_oauth`) it is advisable to add the following `lifecycle` argument to the associated `app_*` resources to prevent the groups being unassigned on subsequent runs:
```

noinarisak (changes requested):
Thanks, @ssttgg for the PR! If you can address my RC the PR is ready to go! 🎉

website/docs/r/app_group_assignment.html.markdown
ssttgg: Good call, done!

ssttgg:
Feedback addressed, ready to go!

noinarisak (approved):

noinarisak:
Thanks, @ssttgg !

PR #182, opened 2020-09-16
Upgrade to terraform-plugin-sdk SDK V2

edulop91:
Upgrading the provider to use the v2 of the terraform plugin SDK.
I am doing this to unlock the ctx functionality in the new sdk to allow me to better track down some crippling okta api rate limiting errors.

I've built the forked version of the provider and seems to be working so far.

Some issues along the way:
- Seems like the use of `Partial` and `SetPartial` are [deprecated](https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html#removal-of-helper-schema-resourcedata-setpartial) and removed.
- Stricter validation means that schemas of `TypeMap` were failing, I've moved them to `TypeSet`
- There were some resources that had a default value and were required. This doesn't make sense and the sdk tests started complaining. Opted for the default value instead of making them required.
- There were some extraneous state functions `StateFunc: normalizeDataJSON,` with the following error: `StateFunc is extraneous, value should just be changed before setting on computed-only field`

I've ignored updating the vendor directory to keep the diff clean but can do so.


noinarisak:
Related to #153

ymylei:
> 
> 
> Related to #153

just an FYI, technically this is different. This is upgrading the terraform plugin sdk while #153 is the okta sdk.

noinarisak:
Closing PR #203 address this now.

PR #187, opened 2020-09-24
Add request throttling to avoid rate limit violations

### LINKED ISSUE #186: Add request throttling to avoid rate limit violations
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Problem

Terraform Okta provider causes rate limit violations. This is problematic for users who have apps that rely on certain Okta APIs. A rate limit violation caused by Terraform Okta provider often effectively causes app downtime. Some organizations that are on the highest pricing plan might have significantly higher rate limits than other organizations on lower pricing plans. Because of that this problem might have gone unnoticed thus far.

Example: `terraform plan` and `terraform apply` cause `/api/v1/apps` API endpoint rate limit violation by making lots of these requests:
1. GET /api/v1/apps/{app_id}/groups?limit=200
2. GET /api/v1/apps/{app_id}/users?limit=200
3. GET /api/v1/apps/{app_id}/groups/{group_id}

We got confirmation from Okta team that all of these endpoints share the same underlying `/api/v1/apps` endpoint rate limit.

Imagine for a moment that your app uses Okta API endpoints that share the `/api/v1/apps` rate limit. When a rate limit violation caused by Terraform Okta provider happens your users start to see errors because some of your app's functionality that relies on `/api/v1/apps` stops working until the rate limit resets.

### Proposal

I suggest we add a new provider config argument https://registry.terraform.io/providers/oktadeveloper/okta/latest/docs#argument-reference:

> `max_requests` - (Optional) (Experimental) controls how many requests can be made to each Okta endpoint by the provider. It's used to prevent rate limit violations. By default request throttling is disabled meaning provider might cause rate limits violations. Expects an integer representing a percentage value - e.g. `40`. `40` means that provider is allowed to use up to 40% of the rate limit. E.g. assuming rate limit for `/api/v1/apps` endpoint is 25, up to 10 requests will be made that burn `/api/v1/apps` rate limit. Currently request throttling works only for `/api/v1/apps` rate limit.

I'd mark this feature as "Experimental" for 2 months to get some feedback from early adopters. I know for sure my organization will start using this feature on day one.

Accepting a signle percentage value is the easiest feature to implement. A more advanced config could look like this:
```
{
	"/api/v1/apps": 40,
	"/api/v1/apps/{id}": 100,
	"/api/v1/groups/{id}": 80
}
```

### Solution

I propose we start with only supporting throttling of requests that burn the `/api/v1/apps` rate limit. It's by far the most pressing problem because `/api/v1/apps` rate limit is significantly lower than any other endpoint.

I think we should use http client's transport to throttle the requests. I propose we change https://github.com/oktadeveloper/terraform-provider-okta/blob/b563d6cbbb2d9e69f76f2ca2ed0ae5a217f221ef/okta/config.go#L48 to:
```
	httpClient.Transport = logging.NewTransport("Okta", httpClient.Transport)
	if c.max_requests < 100 {
		httpClient.Transport = NewRequestThrottleTransport(httpClient.Transport, c.max_requests)
	}
```
RequestThrottleTransport pseudocode:
```
type RequestThrottleTransport struct {
	base http.RoundTripper
	maxRequests int
}
func (throttleTransport *RequestThrottleTransport) RoundTrip(req *http.Request) (res *http.Response, err error) {
	// Check if req uses one of the supported endpoints that we know how throttle.
	// Check how many requests were made during current rate limit interval.
	// Do time.Sleep(...) before making the request if needed.
	resp, err := throttleTransport.base.RoundTrip(req)
	// check response headers, save relevant information
	return resp, err
}
```

Please check https://github.com/oktadeveloper/terraform-provider-okta/pull/187

### Problem with rate limit docs

https://developer.okta.com/docs/reference/rl-global-mgmt/ rate limit documentation is really confusing. By reading the docs you'd never guess that these requests:
1. GET /api/v1/apps/{app_id}/groups?limit=200
2. GET /api/v1/apps/{app_id}/users?limit=200
3. GET /api/v1/apps/{app_id}/groups/{group_id}

share the same rate limit with `/api/v1/apps` endpoint. It will be very hard to implement correct rate limit handling of rate limits other than `/api/v1/apps` for community members that aren't Okta employees.

### FAQ
1. > Does `backoff`, `min_wait_seconds` and `max_wait_seconds` solve your issue?

    No because these settings control the provider's behavior after the rate limit violation happened. We want to avoid rate limit violation from happening at all cost.
2. > Can't you use `-parellelism=1` as in `terraform apply -parellelism=1` to mitigate this issue?

    Even with `-parellelism=1` enough requests are made to Okta to cause a rate limit violation.
3. > What about `sleep()`? Does that help?

    sleep will have no inpact on the `plan` phase which causes the rate limit violations.
4. > Is there any workaround that Terraform Okta provider users can use right now?

    You have to either maintain a custom fork or use a proxy that slows down requests made to Okta. With proxy you have to be really careful to not cause timeouts and if you are a Terraform Cloud user you can't use the proxy workaround.

5. > Latest Golang Okta SDK offers request retries on 429 response status code. Surely that solves your problem, right?

    It doesn't help us at all. We can't afford to have the rate limit completely exhausted by terraform-provider-okta. If terraform-provider-okta uses the whole rate limit allowance our app will become unusable until the rate limit resets. We want to tell terraform-provider-okta to only use up to 25% of the rate limit and leave the remaining 75% for our app.

adw1n:
Closes https://github.com/oktadeveloper/terraform-provider-okta/issues/186

I submitted the CLA on 20200804.

Co-authored-by: @programmer04


bogdanprodan-okta:
Hi @adw1n! Thanks for submitting this PR, however, I'm closing this PR as the latest Golang Okta SDK has this feature out of the box.

adw1n:
@bogdanprodan-okta you are very much mistaken. You are probably referring to https://github.com/okta/okta-sdk-golang/tree/v2.2.1#connection-retry--rate-limiting which was added as part of 2.1.0 release.

I go into detail in the issue description **why request retries on rate limit violations don't help at all.** The point of this PR is to **avoid rate limit violations in the first place because rate limit violations cause downtime of Okta customers' apps**. Even if somehow okta-sdk-golang had the feature (I'm confident that it doesn't) that we, the customers are looking for I still don't see any code in terraform-provider-okta that would let us configure maximum number of requests made to Okta by terraform-provider-okta within one rate limit interval.

@bogdanprodan-okta @noinarisak please can we have this PR reopened? Please check the issue and see that many community members have upvoted :+1: the issue and that it's a real problem that prevents many people from using terraform-provider-okta.

bogdanprodan-okta:
Hi @adw1n ! Thanks for the reply! I've checked toy PR one more time and it seems a good-to-have feature to avoid rate limiting errors. Request throttling to avoid rate limits is a more general feature for Okta API, so it would be great to have it not only here. I will discuss the possibility to include this feature in `okta-sdk-golang` instead in the form of `ConfigSetter`, with `max_requests` config option.

adw1n:
> will discuss the possibility to include this feature in okta-sdk-golang instead in the form of ConfigSetter

@bogdanprodan-okta until we have that in okta-sdk-golang is there a problem with having request throttling done in terraform-provider-okta? The feature that you have in mind might not be added for months/years or it might never be added. This pull request is almost ready, the code is certainly good enough to be used and it unblocks a lot of people. Moreover even if you had a new `ConfigSetter` in okta-sdk-golang you'd still need some of the changes done in this pull request to allow people to start using request throttling.

Please notice that **we want to throttle per rate limit instead of per endpoint.** There is a huge difference between the two because **many endpoints share the same rate limit.** I honestly doubt that you can add a `ConfigSetter` that works in a per rate limit mode in a timely fashion.

@bogdanprodan-okta @noinarisak @bretterer please can we have this PR reopened?

adw1n:
> I will discuss the possibility to include this feature in okta-sdk-golang instead

@bogdanprodan-okta  in my opinion the right thing to do is first to have the discussion, then to agree on something, next share the results of the discussion and only then possibly close the PR. I think that until the discussion had been held the PR shouldn't be closed.

bogdanprodan-okta:
Hi @adw1n ! I have a discussion about adding this feature to the terraform provider and I have a green light on that. However, I have some comments about the implementation. I can write some suggestions here (like error handling, adding context, etc.) or just commit to this PR.

adw1n:
@bogdanprodan-okta this is great news!

> However, I have some comments about the implementation. I can write some suggestions here (like error handling, adding context, etc.) or just commit to this PR.

I wouldn't expect anything less. Please feel free to do whichever you prefer - all help is welcome. I should have time this weekend to respond to comments or to do further testing.

okta/provider.go
```diff
@@ -81,39 +81,45 @@ func Provider() terraform.ResourceProvider {
...
-				Optional:    true,
-				Default:     30,
-				Description: "minimum seconds to wait when rate limit is hit. We use exponential backoffs when backoff is enabled.",
-			},
 			"max_wait_seconds": {
 				Type:        schema.TypeInt,
 				Optional:    true,
 				Default:     300,
-				Description: "maximum seconds to wait when rate limit is hit. We use exponential backoffs when backoff is enabled.",
+				Description: "Maximum seconds to wait when rate limit is hit. We use exponential backoffs when backoff is enabled.",
```
programmer04: @bogdanprodan-okta Is this description correct? It seems to me that it sets just request timeout for SDK client...

bogdanprodan-okta: yeah, the description is correct, but the value is misused. It should be:
```okta.WithRateLimitMaxBackOff(int64(c.maxWait)),```
instead of
```okta.WithRequestTimeout(int64(c.maxWait))```
As for request timeout, it should be either `0` or there should be an additional config introduced.

Comment (programmer04) [reply okta/provider.go]: Thanks for the very fast answer! I introduced `request_timeout` and fixed `max_wait_seconds` properties

**Review by bogdanprodan-okta — APPROVED** [review]

**Comment by bogdanprodan-okta** [comment]
LGTM. I will merge this PR shortly, but just in case I'll make some local testing to be sure that everything works as expected

**Comment by bogdanprodan-okta** [comment]
I'm still testing the solution, cause it's not working as expected. Because the code does not know about what is current rate limit, it should receive at least one response from the endpoint to get the number. Until then HTTP client allows doing the requests until the first response and there is a chance of receiving 429 anyway. 
This can be verified using this code snippet:
```go
        httpClient := cleanhttp.DefaultClient()
	httpClient.Transport = newRequestThrottleTransport(httpClient.Transport, 80)
	_, client, err := okta.NewClient(
		context.Background(),
		okta.WithOrgUrl("https://my_company.oktapreview.com"),
		okta.WithToken("my_token"),
		okta.WithCache(false),
		okta.WithHttpClient(*httpClient),
		okta.WithRateLimitMaxBackOff(int64(360)),
		okta.WithRequestTimeout(int64(360)),
		okta.WithRateLimitMaxRetries(int32(10)),
		okta.WithUserAgentExtra("okta-terraform/3.6.1"),
	)
	if err != nil {
		panic(err)
	}
	wg := new(sync.WaitGroup)
	for i := 0; i < 100; i++ {
		go func(c int) {
			wg.Add(1)
			defer wg.Done()
			_, _, err := client.Application.ListApplications(context.Background(), nil)
			if err != nil {
				fmt.Println(err)
				return
			}
		}(i)
	}
	wg.Wait()
```
But because Okta HTTP client uses retries on 429, it still works without errors. I will try to come up with a solution and change the behavior of the code. 

Feel free to write any suggestions here! Thanks!

adw1n:
@bogdanprodan-okta  you're right. According to https://developer.okta.com/docs/reference/rl-global-mgmt/ the smallest rate limit in all pricing models is 20 so we could use `rateLimit=20` until we get the first response. Or we could wait for the first response before we allow any more requests.

adw1n:
The rate limit reset detection could use some improvements before we merge this. I'll have time to think about it this Sunday.

Problematic scenarios:
## 1
12:00:00 request `#1` to /api/v1/apps
12:00:01 response `#1`, state: rateLimitResetTime=12:01:00
12:01:58 request `#2` to /api/v1/apps
12:01:58 request `#3` to /api/v1/apps
12:01:58 request `#4` to /api/v1/apps
12:01:58 request `#5` to /api/v1/apps
12:01:58 request `#6` to /api/v1/apps, sleep for 0 seconds (12:01:58 - 12:01:00)

## 2
12:00:00 request `#1` to /api/v1/apps
12:00:01 response `#1`, state: rateLimitResetTime=12:01:00
12:01:03 request `#2` to /api/v1/apps, state: noOfRequestsMade=2
12:01:04 request `#3` to /api/v1/apps, state: noOfRequestsMade=3
12:01:05 request `#4` to /api/v1/apps, state: noOfRequestsMade=4
12:01:06 request `#5` to /api/v1/apps, state: noOfRequestsMade=5
12:01:07 response `#2`, state: noOfRequestsMade=1 (correct value is 4)

eparisca:
@adw1n @bogdanprodan-okta Is something holding this PR back?

adw1n:
@eparisca afaik this is the last piece of the puzzle https://github.com/oktadeveloper/terraform-provider-okta/pull/187#issuecomment-740608796 that we should probably improve before merging this PR. All help is welcome.

eparisca:
Is there a way to handle (catch) rate limit errors and wait until "reset time" before retrying the request?

adw1n:
@eparisca sure but that doesn't help us at all in this PR. We want to avoid causing rate limit errors.

bogdanprodan-okta:
> @eparisca sure but that doesn't help us at all in this PR. We want to avoid causing rate limit errors.

Currently, there are discussions inside Okta around rate-limiting and there are several ways to solve the problem (like making a `HEAD` request to the endpoint, which won't affect the rate limit but will return headers about them), but I need to ask where they are on solving this.

Another problem is that terraform can run staff concurrently, i.e. running the provider's binary in separate threads. This causes the problem of sharing the rate-limit state between these threads. I think I'll have to investigate whether it's possible to use terraform's state file as a sort of cache for rate-limit value. If that's not possible, then I'll try to find another way.

eparisca:
If we know the rate limit ceiling (after the first request), if we reach the limit, there is nothing to do. But the majority of the cases will not be in this corner case. Can we set that aside?  

If we know the rate limit ceiling (after the first request), can we calculate how close we are to the limit (threshold) and make a decision on whether to delay the request? 

Have you already considered the following client-side strategies?   
* https://cloud.google.com/solutions/rate-limiting-strategies-techniques#client-side_strategies
* https://builtin.com/software-engineering-perspectives/pre-emptive-client-side-rate-limiting

eparisca:
> Another problem is that terraform can run staff concurrently, i.e. running the provider's binary in separate threads. This causes the problem of sharing the rate-limit state between these threads. I think I'll have to investigate whether it's possible to use terraform's state file as a sort of cache for rate-limit value. If that's not possible, then I'll try to find another way.

@bogdanprodan-okta We are currently experimenting with the Terraform CLI `parallelism` argument on our Okta resources to reduce the risk of reaching the limits.

bogdanprodan-okta:
> If we know the rate limit ceiling (after the first request), if we reach the limit, there is nothing to do. But the majority of the cases will not be in this corner case. Can we set that aside?
> 
> If we know the rate limit ceiling (after the first request), can we calculate how close we are to the limit (threshold) and make a decision on whether to delay the request?
> 
> Have you already considered the following client-side strategies?
> 
> * https://cloud.google.com/solutions/rate-limiting-strategies-techniques#client-side_strategies
> * https://builtin.com/software-engineering-perspectives/pre-emptive-client-side-rate-limiting

This PR basically does what you just wrote. But the problem with concurrency is that calls are independent. Let say the provider makes a call, calculates that it should wait 5 seconds, then makes another call and gets 429 error. Why? Because there were 10 concurrent calls in 10 different threads and they have no idea about each other, so during this 5-sec wait, other threads drained the rate-limit. 

Another problem is the provider can not pre-calculate wait time. It should make a call and then decide how much time it should wait. So there might be a situation when the rate-limit was affected from somewhere else, so the response will be 429 regardless. That's why there should be some mechanism to make a pre-call to figure out the current rate-limit in order to proceed. But again, there should be some stateful resource for that.

As I was saying earlier, there should be some other solution to solve this problem.

monde:
I'm closing this pull request. We used the techniques @adw1n demonstrated here for a percentage based approach that can be set with `max_api_capacity` value in the provider settings. This was implemented in #556 . In this example the provider will only use up 75% of API capacity allotment of any of the metered (see https://developer.okta.com/docs/reference/rl-global-mgmt/ ) management endpoints it utilizes (currently aps ->`/api/v1/apps`, users -> `/api/v1/users`, other -> `/api/v1/xxxxxx`.

```
provider "okta" {
  org_name  = "dev-123456"
  base_url  = "oktapreview.com"
  api_token = "xxxx"
  max_api_capacity = 75
}
```

The `max_api_capacity` will be available in the next release.

PR #190, opened 2020-09-30
Remove requirement for priority in resource_okta_auth_server_policy

aguynamedjoetoo:
According to the policy API object, priority is not required: https://developer.okta.com/docs/reference/api/policy/#policy-object

This change just simply switches this to optional so you don't need to specify priority which often shows as changed in terraform is objects are added or removed.

I tried to dig around to see if anything was needed in the SDK layer and such but didn't appear so, let me know if I missed something!

### DIFF
--- a/okta/resource_okta_auth_server_policy.go
+++ b/okta/resource_okta_auth_server_policy.go
@@ -34,7 +34,7 @@ func resourceAuthServerPolicy() *schema.Resource {
 			"status": statusSchema,
 			"priority": &schema.Schema{
 				Type:        schema.TypeInt,
-				Required:    true,
+				Optional:    true,
 				Description: "Priority of the auth server policy",
 			},
 			"description": &schema.Schema{
--- a/website/docs/r/auth_server_policy.html.markdown
+++ b/website/docs/r/auth_server_policy.html.markdown
@@ -35,7 +35,7 @@ The following arguments are supported:
 
 * `status` - (Optional) The status of the Auth Server Policy.
 
-* `priority` - (Required) The priority of the Auth Server Policy.
+* `priority` - (Optional) The priority of the Auth Server Policy.
 
 * `description` - (Optional) The description of the Auth Server Policy.
 


noinarisak:
Hi, @Joeskyyy thanks for the PR. 

I looked into the code and even though documentation on the policy-object states its `Required=NO` (I think this documentation is wrong, but to need to investigate deeper) the actional implementation API call does require it on CREATE and UPDATE, I'm basing it off this documentation: https://developer.okta.com/docs/reference/api/authorization-servers/#policy-properties.

aguynamedjoetoo:
Thanks @noinarisak -- Some of our existing scripts call the policy-object directly and don't specify the priority, what appears to happen is the API just inserts it as a incremental from the last found one (e.g. if I have 49, the next call creates a new policy with priority 50).

Totally understanding wanting to triple check though, let me know if there is anything I can contribute on. This is a blocker for us building some terraform modules to support this being in terraform at scale, just as a heads up, without making some ugly workarounds.

noinarisak:
Hi, @Joeskyyy thanks for the additional details. I submitted an issue on our documentation repo at [issue 1541](https://github.com/okta/okta-developer-docs/issues/1541) for additional clarification and now there is an internal Jira ticket but no assignment just yet. 🤞 

Sadly, your workaround will have to be in place untill we get clarification. With so many endpoints and ongoing community contribution, I hate to add confusion between our documentation and what has been implemented in the provider. I hope you understand. 😞

aguynamedjoetoo:
@noinarisak Totally understand, no need to break 329462938798 other people by accident without doing proper checks (:

bogdanprodan-okta:
Hi @Joeskyyy ! I've looked closely into this property and it turned out to be required because this is the policy for authorization servers which is slightly different than the others. You can check documentation [here](https://developer.okta.com/docs/reference/api/authorization-servers/#policy-properties).

aguynamedjoetoo:
Thanks for the followup @bogdanprodan-okta, given that do you have any suggestion as to how one should handle creating/importing existing resources? The KISS is obviously just monotonically incrementing, but that is super flat and doesn't really allow for as nice of a modular approach

bryantbiggs:
@Joeskyyy if you import then run a plan, it should tell you what it currently is and what its trying to change to. just update to what the current value is (import + update code to align with existing => nearly empty diff)

aguynamedjoetoo:
@bryantbiggs sure but that's just less than ideal to just always have to +1 and go and manually update a bunch of priorities. The experience in the console is much different, which is why trying to automate this has come as a bit of a shocker.

bryantbiggs:
I mean, thats the nature of Terraform and IaC - its very explicit. and thats not just Okta's provider, its Terraform's core principle. you can try setting the priority as null, this sometimes works for cases like this where you don't care to control the value as much

aguynamedjoetoo:
@bryantbiggs whoops sorry some how missed your comment -- Sadly null doesn't work :\ I of course understand it's part of terraform's core, just seems like the okta API should handle the incrementing should the priority not be sent with the request, but I understand there's likely a lot more behind the curtains. I'll close this out

ryber:
It would be great if SOMETHING could be done with priority in the Okta API. We have a fairly complex project with Policies for different scopes in different files. It's not at all convenient or easy to keep track of which priority is which and it honestly doesn't matter. So we just live with terraform saying it's going to modify 80 different things every time. I would love to just leave it out entirely and let Okta append the policy.

PR #195, opened 2020-10-09
Renaming Go module as per the organisation move

stack72:
Pure mechanical changes


noinarisak:
Thanks, @stack72! For catching that!

PR #206, opened 2020-10-29
Fix population of the user 'status' attribute

isometry:
The `status` attribute on the `okta_user` and `okta_users` data providers was not set; this patch corrects that.

### DIFF
--- a/okta/user.go
+++ b/okta/user.go
@@ -425,6 +425,8 @@ func flattenUser(u *okta.User, d *schema.ResourceData) (map[string]interface{}, 
 		}
 	}
 
+	attrs["status"] = mapStatus(u.Status)
+
 	data, err := json.Marshal(customAttributes)
 	if err != nil {
 		return nil, fmt.Errorf("Failed to load custom_attributes to JSON")


bogdanprodan-okta:
Hi @isometry! Thanks for the PR! Regarding the user `status` attribute, I think status is set correctly [resource_okta_user](https://github.com/oktadeveloper/terraform-provider-okta/blob/master/okta/resource_okta_user.go#L389). If this is not the case you are trying to fix, could you please provide a terraform sample or error/log message where the status is not being set correctly?

isometry:
@bogdanprodan-okta , the issue is with the data sources rather than the resource. The documentation for both `data okta_user` and `data okta_users` claims that they return current user status in the `status` attribute, but they actually just return an empty string in the attribute because it isn't initialised properly. This PR fixes that, returning the current user status rather than a blank string.

Example:

```hcl
data okta_users example {
  search {
    name       = "profile.login"
    comparison = "sw"
    value      = ""
  }
}
```

_Without this patch_, every user in the returned `data.okta_users.example.users` attribute has `status = ""`.

bogdanprodan-okta:
Hi @isometry! Seems like status isn't populated, however, I think status should be set as in user resource `_ = d.Set("status", mapStatus(user.Status))`

isometry:
Updated patch should do the job then :-)

PR #217, opened 2020-11-13
Adding type to data okta group

dangoslen:
Re-creating this PR from #155 since I royally screwed up a merge. I believe I have merged this much better now and should be only additions for the functionality I'm attempting to implement.

---
Description for original PR:

This PR adds the ability to filter by a type in the okta_group data entity.

This is useful in cases were Okta is managed groups from multiple sources (LDAP, Google groups, etc.) that have the same name but different types.

It additionally returns the type as an attribute of the entity.


tgoodsell-tempus:
https://github.com/oktadeveloper/terraform-provider-okta/blob/7231dbe0ca30154c8f40508da6b51c95091f9fb3/okta/data_source_okta_group.go#L63

This should be changed to optionally include the type if used, maybe something like:
```
if err != nil {
		return fmt.Errorf("failed to query for groups: %v", err)
	} else if len(groups) < 1 {
                if d.Get("type") != nil {
                    return fmt.Errorf("Group \"%s\" not found with type \"%s\"", name, d.Get("type").(string))
               } else {
	           return fmt.Errorf("Group \"%s\" not found", name)
              }
	}
```

examples/okta_group/datasource.tf
```diff
@@ -1,17 +1,23 @@
-resource okta_group test {
+resource "okta_group" "test" {
```
bogdanprodan-okta: No need to set quotes here.

examples/okta_group/datasource.tf
```diff
@@ -1,17 +1,23 @@
...
+data "okta_group" "test" {
   include_users = true
   name          = "${okta_group.test.name}"
 }
+
+data "okta_group" "test_type" {
+  include_users = true
+  name          = "${okta_group.test.name}"
+  type          = "nonExistentType"
+}
```
bogdanprodan-okta: Add newline at the end of the file

examples/okta_group/datasource.tf
```diff
@@ -1,17 +1,23 @@
...
-data okta_group test {
+data "okta_group" "test" {
   include_users = true
   name          = "${okta_group.test.name}"
 }
+
+data "okta_group" "test_type" {
+  include_users = true
+  name          = "${okta_group.test.name}"
+  type          = "nonExistentType"
```
bogdanprodan-okta: The test will fail because you've added invalid group type
```
testing.go:705: Step 0 error: config is invalid: expected type to be one of [OKTA_GROUP APP_GROUP BUILT_IN], got nonExistentType
```

okta/data_source_okta_group_test.go
```diff
@@ -22,10 +24,25 @@ func TestAccOktaDataSourceGroup_read(t *testing.T) {
...
 					resource.TestCheckResourceAttrSet("data.okta_group.test", "id"),
+					resource.TestCheckResourceAttrSet("data.okta_group.test", "type"),
 					resource.TestCheckResourceAttrSet("okta_group.test", "id"),
 					resource.TestCheckResourceAttr("okta_group.test", "users.#", "1"),
 				),
 			},
+			{
+				Config: config,
+				Check:  resource.ComposeTestCheckFunc(testDoesNotExist("okta_group.test_type")),
+			},
```
bogdanprodan-okta, line 35: I think it should be `data.okta_group.test_type`

okta/data_source_okta_group.go
```diff
@@ -44,15 +51,26 @@ func dataSourceGroupRead(d *schema.ResourceData, m interface{}) error {
 
 func findGroup(name string, d *schema.ResourceData, m interface{}) error {
 	client := getOktaClientFromMetadata(m)
-	groups, _, err := client.Group.ListGroups(context.Background(), &query.Params{Q: name})
+	searchParams := &query.Params{Q: name}
+	if d.Get("type") != nil {
```
bogdanprodan-okta: This will never work, as it should be
```
if d.Get("type") != nil && d.Get("type").(string) != ""{
	searchParams.Filter = fmt.Sprintf("type+eq+\"%s\"", d.Get("type").(string))
}
```
https://developer.okta.com/docs/reference/api/groups/#list-groups-with-type

okta/data_source_okta_group.go
```diff
@@ -18,6 +18,13 @@ func dataSourceGroup() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 			},
+			"type": &schema.Schema{
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      false,
```
bogdanprodan-okta: Default is `interface{}`, so it will set `0` as a default group type

dangoslen:
@tgoodsell-tempus mind taking another look at this? TY

tgoodsell-tempus (approved):
LGTM!

bogdanprodan-okta (changes requested):

examples/okta_group/datasource.tf
dangoslen: Ahh yes. This was around before adding that validation. Will update to look for the wrong group type

okta/data_source_okta_group.go
dangoslen: 🤔 I'm trying to read up on where that comes from - can you give me an article or documentation reference to look at?

bogdanprodan-okta: I think this has something to do with how the terraform converts values from the interface into the value. Because type is `schema.TypeString`, I think it just represents false as 0 (because false internally is just integer) and then converts it into the string (using `strings.Itoa` I guess), which is "0".

PR #228, opened 2020-12-09
Overhaul idp_discovery_rule documentation

eatplaysleep:
Documentation was incredibly inaccurate. Updating based on deciphering actual implementation + Okta documentation.


bryantbiggs:
@eatplaysleep would you mind resolving the conflicts

website/docs/r/policy_rule_idp_discovery.html.markdown
```diff
@@ -34,20 +61,71 @@ resource "okta_policy_rule_idp_discovery" "example" {
 ## Argument Reference
 
 The following arguments are supported:
+* `policy_id` - (Required) Policy ID.
```
bogdanprodan-okta: Based on the code, it's `policyid`. There are two options here: either change from `policyid` to `policy_id` in the code as well or add just `policy_id` to the code and mark `policyid` as deprecated and `ConflictsWith:[]string{"policy_id"},`

eatplaysleep: Ah, forgot to change that. I noticed that in the code (with a comment). I'll update the documentation to match the code (for now).

PR #250, opened 2020-12-19
Invalid Rule Handling

ymylei:
Designed to deal with #157 

A group rule becomes `INVALID` when one of the groups in the `group_assignments` is deleted (generally via the API) while the rule itself is still using it.

A group rule in the `INVALID` state can only be read or deleted, no other operations appear to be supported.

This adds handling to the provider to allow it to recognize this state, and handle it so when it is encountered, it will allow the rule to still be deleted, and additionally uses the customdiff to ensure that


ymylei:
I could use additional assistance with this one, since the "free" okta accounts do not have the ability to create / use group rules.

Mainly someone who can run the acceptance tests and see if they're working and as-desired.

bogdanprodan-okta:
Hi @ymylei! Thanks for the PR! I'll run the acceptance tests using test org. If everything is good, I'll merge the PR!

PR #154, opened 2020-07-20
Add support for group target for administrative roles

jlaundry:
- Renames **okta_group_roles** to **okta_group_role** as recommended by @rtlong in issue #13
- Provides a new **okta_group_role_group_target** resource

This partially addresses issue #13, by providing the first half of that issue - app targets could be added as another resource, I just don't have scope to develop and test that right now.

This is my first TF module, so please highlight any style/convention issues 😊

PR #255, opened 2020-12-24
New Resource: okta_group_role

ymylei:
This PR creates a new resource: `okta_group_role`

This resource is designed to replace `okta_group_roles` such that:

- Manages the admin role assignments at an individual level, instead of the list of roles that `okta_group_roles` manages with.
- This allows the ID for the resource to be exactly the `roleID` defined in the API docs

This PR is part 1 of a few PRs for resources I will release, the others being:

1. Group target list for admin roles
2. App instance target list for admin roles
3. App catalog/name target list for admin roles

This resource will provide the "base" that I will use to work on the others (although technically it's not required). In this case having the individual role assignment ID is a requirement for using the API endpoints for the target lists, so this would be immensely more helpful than how `roles` works.

This is intended to replace #154  with the additional caveats of:

- Not modifying the `okta_group_roles` resource, only adding a deprecation message. This should allow us to avoid a breaking change and ease people into the new pattern. (Follows best practice per the hashicorp docs)
- Appending the various admin role scoping functionality into this resource (or standalone resources where possible).

### DIFF
--- a/examples/okta_group_role/README.md
+++ b/examples/okta_group_role/README.md
@@ -1,0 +1,6 @@
+# okta_group_role
+
+Represents an assignment of an Admin role to an Okta
+Group. [See Okta documentation for more details](https://developer.okta.com/docs/reference/api/roles/#list-roles-assigned-to-group)
+
+- Example of a group assigned as a `READ_ONLY_ADMIN` [can be found here](./basic.tf)
--- a/examples/okta_group_role/basic.tf
+++ b/examples/okta_group_role/basic.tf
@@ -1,0 +1,20 @@
+// Test group & user assigned to group
+resource "okta_group" "test" {
+  name = "testAcc_replace_with_uuid"
+  description = "testing"
+}
+
+resource "okta_user" "test" {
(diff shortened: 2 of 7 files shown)


ymylei:
Personal test results:
```
tom@DESKTOP-NNSTVLT:~/terraform-provider-okta$ make testacc TEST=./okta TESTARGS='-run=TestAccOktaGroupAdminRole_crud'
go mod tidy
go mod vendor
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./okta -v -run=TestAccOktaGroupAdminRole_crud  -timeout 120m
=== RUN   TestAccOktaGroupAdminRole_crud
--- PASS: TestAccOktaGroupAdminRole_crud (7.12s)
PASS
ok      github.com/oktadeveloper/terraform-provider-okta/okta   7.125s
```

bogdanprodan-okta:
Hi, @ymylei! Thanks for the PR! Everything looks good, I just have to double-check everything and I think this might be included in the upcoming 3.8.0 release.

PR #335, opened 2021-02-18
Okta Group Membership - Wait after create

ymylei:
This is a primitive bug fix for the issue I've encountered, documented below.

There can be a lag upon group membership addition operation, which does not mesh well with the ability to perform the read function after create. Therefore, my simple fix is to set a wait/delay between the addition operation and the read operation (only for creates); since there is no ID or "404-like" error we can reliably use as a "re-try" trigger.

```
    apply.go:71: 

        	Error Trace:	apply.go:71

        	            				module_test.go:22

        	Error:      	Received unexpected error:

        	            	FatalError{Underlying: error while running command: exit status 1; 

        	            	Error: Provider produced inconsistent result after apply

        	            	

        	            	When applying changes to

        	            	module.test_machine_user_okta_admin.okta_group_membership.group_membership["Test_Dummy_Group"],

        	            	provider "registry.terraform.io/oktadeveloper/okta" produced an unexpected new

        	            	value: Root resource was present, but now absent.

        	            	

        	            	This is a bug in the provider, which should be reported in the provider's own

        	            	issue tracker.

        	            	}

        	Test:       	TestMachineUserOktaAdminV2

```

### DIFF
--- a/okta/resource_okta_group_membership.go
+++ b/okta/resource_okta_group_membership.go
@@ -3,6 +3,7 @@ package okta
 import (
 	"context"
 	"fmt"
+	"github.com/cenkalti/backoff/v4"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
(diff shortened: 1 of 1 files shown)


bogdanprodan-okta:
Hi, @ymylei! Thanks for the PR! I don't like having 'static' sleep in the code, so I think it better to implement a sort of `read with exponential retry` instead of just calling `return resourceGroupMembershipRead(ctx, d, m)`.
Basically, it's better to do something like:
```
func resourceGroupMembershipCreate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
	groupId := d.Get("group_id").(string)
	userId := d.Get("user_id").(string)
	logger(m).Info("adding user to group", "group", groupId, "user", userId)
	_, err := client.Group.AddUserToGroup(ctx, groupId, userId)
	if err != nil {
		return diag.Errorf("failed to add user to group: %v", err)
	}
	d.SetId(fmt.Sprintf("%s+%s", groupId, userId))
	bOff := backoff.NewExponentialBackOff()
	bOff.MaxElapsedTime = time.Second*5
	bOff.InitialInterval = time.Second
	err = backoff.Retry(func() error {
		inGroup, err := checkIfUserInGroup(ctx, client, groupId, userId)
		if err != nil {
			return backoff.Permanent(fmt.Errorf("unable to complete group check for user: %v", err))
		}
		if inGroup {
			return nil
		}
		return errors.New("user is not in group")
	}, bOff)
	if err != nil {
		var pErr backoff.PermanentError
		if errors.Is(err, &pErr) {
			return diag.FromErr(err)
		}
		d.SetId("")
	}
	return nil
}
```

What do you think?

ymylei:
@bogdanprodan-okta Your idea makes a ton of sense, didn't realize that not using the `read` function wasn't actually a requirement of terraform. I've gone ahead and shamelessly copied your example, with a bit of trimming in the error handling, since it looks like it would be unnecessary for us to check if the error was permanent or not.

bogdanprodan-okta:
> @bogdanprodan-okta Your idea makes a ton of sense, didn't realize that not using the `read` function wasn't actually a requirement of terraform. I've gone ahead and shamelessly copied your example, with a bit of trimming in the error handling, since it looks like it would be unnecessary for us to check if the error was permanent or not.

The reason there is read after create/update is that sometimes some fields are computed, and some are default. To be sure that all the fields are exactly the same after the update and create operations, the read is performed.

Your PR looks good, so gonna merge it. I will release this fix by the end of next week with a bunch of other stuff. Thanks!

Issue #237, opened 2020-12-15
Doesnt appear to be a way to add scope to an application


mcrobbj:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Affected Resource(s)

- okta_app_oauth

### Terraform Configuration Files

```hcl
resource "okta_app_oauth" "middleware" {
  label          = "middleware"
  type           = "service"
  response_types = ["token"]
  grant_types    = ["client_credentials"]
  token_endpoint_auth_method = "private_key_jwt"

  jwks {
    kty = "RSA"
    kid = "c08ff170-a666-404d-hhhhhhhh
    e   = "AQAB"
    n   = "bbbbbbbbbbbb"
  }
}
```

### Expected Behavior

I would like to add the okta.manage.groups, okta.manage.user scopes

### Actual Behavior
I cant add scopes to the application

### Steps to Reproduce

1. `terraform apply`

### References

- #0000

bogdanprodan-okta:
Hi @mcrobbj! Thanks for creating this issue! I'll check it shortly and will get back to you asap.

tgoodsell-tempus:
Probably requires a new resource associated with the OAUTH apps?

Reference: https://developer.okta.com/docs/reference/api/apps/#application-oauth-2-0-scope-consent-grant-operations

mcrobbj:
> Probably requires a new resource associated with the OAUTH apps?
> 
> Reference: https://developer.okta.com/docs/reference/api/apps/#application-oauth-2-0-scope-consent-grant-operations

I think so, you also probably need the do the get for existing okta scopes such as okta.users.manage as the api's work off id's rather than names.

CodyDunlap:
Is this a duplicate of #69?

mcrobbj:
Sorry yes it does seem to be a duplicate

mcrobbj:
#356 <https://github.com/oktadeveloper/terraform-provider-okta/pull/356> states
that its fix and it looks like it would be but its not been released.

patrickmarshallcb:
[...]

PR #356, opened 2021-02-26
Add api scopes

### LINKED ISSUE #69: Add support for creating 'Okta API scopes'
*@rachrdfds commented on Dec 2, 2019, 3:01 PM UTC:*

**Is your feature request related to a problem? Please describe.**
When creating an application that is to be used to create and maintain other Okta applications etc. in Okta, some Okta API scopes must be enabled/granted on the application.

Currently I can create the application using Terraform but the I ahve to set the Okta API scopes manually (GUI):
[image](https://user-images.githubusercontent.com/57186882/69969333-856cf780-151c-11ea-8782-30cf63d2aa62.png)

**Describe the solution you'd like**
A way to create the application with the specified Okta API scopes. It could be an array of strings containing one or more of (below is currently a complete list):

*   okta.apps.manage
*   okta.apps.read
*   okta.authorizationServers.manage
*   okta.authorizationServers.read
*   okta.clients.manage
*   okta.clients.read
*   okta.clients.register
*   okta.eventHooks.manage
*   okta.eventHooks.read
*   okta.events.read
*   okta.factors.manage
*   okta.factors.read
*   okta.groups.manage
*   okta.groups.read
*   okta.idps.manage
*   okta.idps.read
*   okta.inlineHooks.manage
*   okta.inlineHooks.read
*   okta.logs.read
*   okta.roles.manage
*   okta.roles.read
*   okta.schemas.manage
*   okta.schemas.read
*   okta.users.manage
*   okta.users.manage.self
*   okta.users.read
*   okta.users.read.self

**Describe alternatives you've considered**
I don't see any workaround for this.

*This issue was moved by [noinarisak](https://github.com/noinarisak) from [articulate/terraform-provider-okta#361](https://github.com/articulate/terraform-provider-okta/issues/361).*

mariussturm:
This PR adds a new resource `okta_app_oauth_api_scope` in order to manage OAuth API scopes.
First-time provider dev here might be that there is some oddity.

Fixes #69, #237


bogdanprodan-okta:
Hi, @mariussturm! Thanks for submitting this PR. I'll review it and get back to asap.

mcrobbj:
Is this going to be released?

bogdanprodan-okta:
> Is this going to be released?

Hi! Yes, this is going to be released by the end of this week.

okta/resource_okta_app_oauth_api_scope.go
```diff
@@ -0,0 +1,242 @@
...
+			"issuer": {
+				Required:    true,
+				Type:        schema.TypeString,
+				Description: "The issuer of your Org Authorization Server, your Org URL.",
+			},
+			"scopes": {
+				Type:        schema.TypeList,
+				Required:    true,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Description: "Scopes of the application for which consent is granted.",
```
bogdanprodan-okta, line 81: Could you please add validation for the scopes based on the list from [here](https://developer.okta.com/docs/guides/implement-oauth-for-okta/scopes/)?

okta/resource_okta_app_oauth_api_scope.go
```diff
@@ -0,0 +1,242 @@
...
+	for _, scope := range d.Get("scopes").([]interface{}) {
+		desiredScopes = append(desiredScopes, scope.(string))
+	}
+
+	// extract scope list form []okta.OAuth2ScopeConsentGrant
+	for _, currentScope := range from {
+		currentScopes = append(currentScopes, currentScope.ScopeId)
+	}
+
+	// find scopes that should not be there
```
bogdanprodan-okta: As an advice, you can use `splitTargets` (okta/resource_okta_admin_role_targets.go:183) func here. Yes, the naming is confusing, so feel free to rename params.

bogdanprodan-okta (changes requested):

okta/resource_okta_app_oauth_api_scope.go
mariussturm: Totally agree, validation would be good here but it looks like tf doesn't like validation on lists. I get the following error:

```
1 error occurred:
	* resource okta_app_oauth_api_scope: scopes: ValidateFunc and
ValidateDiagFunc are not yet supported on lists or sets.
```

I could put the validation in the setter function but that becomes a bit messy, do you have any idea?

okta/resource_okta_app_oauth_api_scope.go
mariussturm: Ah good catch, we have the same code already. I Will change that!

okta/resource_okta_app_oauth_api_scope.go
bogdanprodan-okta: Try putting validation for each element `Elem:        &schema.Schema{Type: schema.TypeString},`

mariussturm: Ha, stupid me, yes that works perfectly. Thanks again!

bogdanprodan-okta:
The last thing I wanted to ask is to merge with the latest master and run `make fmt`. I've introduced a new more advanced [formatting tool](https://github.com/mvdan/gofumpt) instead of standard `gofmt`.

bogdanprodan-okta (approved):

mariussturm:
Thanks for the approval! Did a rebase and `make fmt`.

blakemann:
@bogdanprodan-okta I've got a project that I'm waiting to be able to use this in - now that it's merged, will there be another release being done soon to include it?

bogdanprodan-okta:
Hi, @blakemann ! The next release will probably this week. The only problem is that provider has migrated from `oktadeveloper` to `okta` and I don't know if I will be able to release it with the current `gpg` key. 
There is always a possibility to use [in-house provider ](https://www.terraform.io/docs/cloud/run/install-software.html#in-house-providers) built from source. Basically, you run `go install`, copy the binary into your `terraform.d` dir and run `terraform init`.

mariussturm:
@bogdanprodan-okta thanks for the investigation! A registry deployment of 3.11 would be great so that we can officially make use of that feature.

tmatilai:
Sorry to comment to an old PR, but is there any reason why the `issuer` attribute couldn't be internally fetched from the provider configuration?

On the client side, the provider attributes are not accessible (especially problem on common modules), and there doesn't seem to exist any data source for getting the URL either.

/cc @bogdanprodan-okta

mariussturm:
@tmatilai if the `issuer` would be fetched from the internal configuration it has to be the org URL under all circumstances and all edge cases. I am not sure if this is always true, especially when custom authorization servers come into play.

What exactly is the use-case for reading back the `issuer` URL? Usually, you know the Okta base URL in the tf code and just render it here again?

tmatilai:
> if the `issuer` would be fetched from the internal configuration it has to be the org URL under all circumstances and all edge cases.

If there are edge cases, then the `issuer` attribute could be optional, and default to the provider configuration. 🤷 

> What exactly is the use-case for reading back the `issuer` URL? Usually, you know the Okta base URL in the tf code and just render it here again?

If the provider is passed from a root module to a common module, the common module doesn't know about the org or base URL any more. So now it's required to add separate variables just for the `issue` URL. 

In case of auth server is created, the base URL can actually be parsed from its issuer URL, but that feels extraneous, too.

Btw, thanks @mariussturm for adding the resource in the first place! 🤗

PR #362, opened 2021-03-03
Removed 'ForceNew' in case policy name changes

### LINKED ISSUE #55: Why are policy names ForceNew?
*@quantumew commented on Oct 18, 2019, 8:28 PM UTC:*

I don't see a reason for policy name attributes to be `ForceNew`. I think this is a remnant of the original iteration of this resource that can be removed.

*This issue was moved by [noinarisak](https://github.com/noinarisak) from [articulate/terraform-provider-okta#306](https://github.com/articulate/terraform-provider-okta/issues/306).*

bogdanprodan-okta:
Fix #55


examples/okta_policy_mfa/basic_updated.tf
```diff
@@ -3,7 +3,7 @@ data "okta_group" "all" {
 }
 
 resource "okta_policy_mfa" "test" {
-  name            = "testAcc_replace_with_uuid"
+  name            = "testAcc_replace_with_uuid_new"
```
bretterer, line 6: Just cuious what the reason for this change is

bretterer (approved):
LGTM, Just one question

examples/okta_policy_mfa/basic_updated.tf
bogdanprodan-okta: I needed to make sure that the name will change.

PR #376, opened 2021-03-11
Fixed default auth server and minor refactoring

bogdanprodan-okta:


okta/resource_okta_app_group_assignment_test.go
```diff
@@ -136,7 +136,7 @@ func ensureAppGroupAssignmentExists(name string) resource.TestCheckFunc {
 	}
 }
 
-func ensureAppGroupAssignmentRetained(appName string, groupName string) resource.TestCheckFunc {
+func ensureAppGroupAssignmentRetained(appName, groupName string) resource.TestCheckFunc {
```
bretterer, line 139: Can you tell me why we no longer have this defined as a `string`?

okta/resource_okta_app_oauth_api_scope.go
```diff
@@ -31,30 +31,29 @@ var validScopes = []string{
 	"okta.policies.manage", "okta.policies.read",
 }
 
-func resourceAppOAuthApiScope() *schema.Resource {
+func resourceAppOAuthAPIScope() *schema.Resource {
```
bretterer, line 34: Assuming since this is internal only, changing the func name, this won't cause any BC issues?

okta/resource_okta_app_user_test.go
```diff
@@ -160,7 +160,7 @@ func checkAppUserDestroy(s *terraform.State) error {
 	return nil
 }
 
-func ensureAppUserRetained(appName string, userName string) resource.TestCheckFunc {
+func ensureAppUserRetained(appName, userName string) resource.TestCheckFunc {
```
bretterer, line 163: Same question about `string`

okta/resource_okta_group_role.go
```diff
@@ -307,7 +307,7 @@ func addGroupTargetsToRole(ctx context.Context, client *okta.Client, groupID, ro
 	return nil
 }
 
-func removeGroupTargetsFromRole(ctx context.Context, client *okta.Client, groupID string, roleID string, groupTargets []string) error {
+func removeGroupTargetsFromRole(ctx context.Context, client *okta.Client, groupID, roleID string, groupTargets []string) error {
```
bretterer, line 310: Same question about `string`

bretterer (commented):
Few questions before approving

okta/resource_okta_app_group_assignment_test.go
bogdanprodan-okta: This is basically a linter check that will say 'plz convert `appName string, groupName string` to `appName, groupName string`.

okta/resource_okta_app_oauth_api_scope.go
bogdanprodan-okta: Changing the names of the function doesn't change the provider's behavior.

bretterer (approved):

PR #433, opened 2021-04-20
Added `display_name` field to the `okta_auth_server_scope` resource

### LINKED ISSUE #430: Unable to manage the "display name" when configuring "okta_auth_server_scope"
I've been working with configuring my Okta instance using terraform and the Okta provider. As part of this, I'm adding new scopes to our authorization server.

Using `okta_auth_server_scope`, I'm able to define it as needed except for the **display name** property.

I've been using the documentation here: https://registry.terraform.io/providers/oktadeveloper/okta/latest/docs/resources/auth_server_scope#argument-reference and have not been able to find any specific information for the display name. 

The net result is that the scope gets created with a Name, Description and other attributes but no display name. I'd eventually be using the display name when requesting user consent as part of some of the flows.

I've attached an image.
image

Wondering if this was just a miss.

bogdanprodan-okta:
Fix #430

### DIFF
--- a/examples/okta_auth_server_scope/basic.tf
+++ b/examples/okta_auth_server_scope/basic.tf
@@ -2,6 +2,7 @@ resource "okta_auth_server_scope" "test" {
   consent        = "REQUIRED"
   description    = "test"
   name           = "test:something"
+  display_name   = "test"
   auth_server_id = okta_auth_server.test.id
 }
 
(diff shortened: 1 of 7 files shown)


bretterer (approved):

ht128:
Hi @bogdanprodan-okta,

Thanks so much for getting this in so quick. Can you please advise what is the timeline for this being released?

Many thanks,

HT

bogdanprodan-okta:
Hi, @ht128! I'm not sure yet, because there are no updates from the terraform team. You can still use the provider compiled from the source though. Please read the note in the[ README file](https://github.com/okta/terraform-provider-okta#important-note).

PR #482, opened 2021-05-23
Mark App Users & Groups Schema as Deprecated

ymylei:
These should be fully replaced by using:
Users: `okta_app_user`
Groups: `okta_app_group_assignments` 

In a future breaking change, the intent would be to remove the handling of users/groups from here since it's redundant with the usage of the resources mentioned above. Would save on API calls.

### DIFF
--- a/okta/app.go
+++ b/okta/app.go
@@ -58,12 +58,14 @@ var baseAppSchema = map[string]*schema.Schema{
 		Optional:    true,
 		Elem:        appUserResource,
 		Description: "Users associated with the application",
+		Deprecated:  "The direct configuration of users in this app resource is deprecated, please ensure you are the resource `okta_app_user` for this functionality.",
 	},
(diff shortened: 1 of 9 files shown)


bogdanprodan-okta:
Hi @ymylei! Because we are planning to release 4.x.x version, we are going to remove all the user and groups fields from all the app-related resources, this PR is redundant.

ymylei:
@bogdanprodan-okta I believe that is a mistake on your part. According to the deprecation best practices from Hashi (https://www.terraform.io/docs/extend/best-practices/deprecations.html) we should be releasing a "deprecated" release before you remove anything in a major cut.

bogdanprodan-okta:
@ymylei yeah, good point. I'll reopen this PR, and make a `3.12.0` release next month.

bogdanprodan-okta (approved):

PR #484, opened 2021-05-24
(WIP) Resource: okta_app_user_assignments

ymylei:
Creates a new resource `okta_app_user_assignments` for bulk assignment management for app users.

Details to come

TODO:
1. Need to figure out how to properly suppress the diffs on the embedded `profile` attribute when it's not being set by us.
2. Merge in changes from #477 
3. Look into handling of the `userScope` to ensure we're only reading/controlling "user" assignment.
4. Add docs
5. Add complex tests of the other major app types, particularly those which actually handle the `credentials` passed.

### DIFF
--- a/examples/okta_app_basic_auth/basic_updated.tf
+++ b/examples/okta_app_basic_auth/basic_updated.tf
@@ -1,11 +1,11 @@
 resource "okta_user" "user" {
   admin_roles = [
     "APP_ADMIN",
-    "USER_ADMIN"]
-  first_name  = "TestAcc"
-  last_name   = "blah"
-  login       = "testAcc-replace_with_uuid@example.com"
-  email       = "testAcc-replace_with_uuid@example.com"
+  "USER_ADMIN"]
+  first_name = "TestAcc"
+  last_name  = "blah"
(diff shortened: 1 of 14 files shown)


bogdanprodan-okta:
Hi @ymylei! Are you planing to finish this PR or do you need some help?

ymylei:
> Hi @ymylei! Are you planing to finish this PR or do you need some help?

@bogdanprodan-okta My apologies, I think I could use some help but I've been away working on this for a while. My big sticking point is working with the sub-resource objects in the schema is currently pretty shit from a `terraform-provider-sdk` / `terraform` standpoint, things like getting it to store & diff properly are a PITA.

It may be better to just leave this be and remain using `okta_app_user` until their new SDK is recommended and providers are pushed to start moving over, or whenever the complex object handling isn't so bad.

If you can help figure out how to make the resource store/handle right on the current SDK then we can try to carry on with this.

bogdanprodan-okta:
@ymylei I think you are facing the same problem I had with `okta_app_group_assignments`. I spent a week trying to make `group` objects behave correctly and not causing constant diffs. I'll try to use the same approach here.

PR #556, opened 2021-08-02
Request governor for Okta Management API calls

### LINKED ISSUE #186: Add request throttling to avoid rate limit violations

monde:
@bogdanprodan-okta @adw1n @programmer04 @eparisca @peteryaworsky-okta
This is a rework of PR #187 but using a percentage of the given API capacity for the Okta API management endpoints.

I found how the AWS provider is using a global mutex to protect too many calls going to AWS security groups API and took their approach here. @peteryaworsky-okta suggested we try this capacity percentage governor/throttle approach when we discussed API rate limiting with his team.

I need to bring tests into the PR on the transport as @adw1n original did in #187 but wanted to get this on everyone's radar first.

- https://github.com/hashicorp/terraform-provider-aws
- https://github.com/hashicorp/terraform-provider-aws/blob/main/aws/provider.go#L1576
- https://github.com/hashicorp/terraform-provider-aws/blob/main/aws/internal/mutexkv/mutexkv.go


monde:
I need to touch up 429 retry logic in addition to bringing in http/transport tests.

monde:
> I need to touch up 429 retry logic in addition to bringing in http/transport tests.

governed transport can run on the retry client or the default client, this is good to go as is.

bogdanprodan-okta (approved):

okta/internal/apimutex/apimutex.go
```diff
@@ -0,0 +1,132 @@
...
+
+	// if the status hasn't been updated recently assume there is capacity
+	now := time.Now().Unix()
+	if status.reset+60 < now {
+		return true
+	}
+
+	// calculate utilization
+	utilization := 100.0 * (float32(status.limit-status.remaining) / float32(status.limit))
+	fmt.Println(utilization)
```
monde: I must have left that while println debugging, will delete

okta/internal/apimutex/apimutex.go
```diff
@@ -0,0 +1,132 @@
...
+		result = "apps"
+	case strings.HasPrefix(endPoint, "/api/v1/users"):
+		result = "users"
+	default:
+		result = "other"
+	}
+	return result
+}
+
+func (s *ApiStatus) Reset() int64 {
```
monde, line 117: need to document public methods

okta/internal/transport/governed_transport.go
```diff
@@ -0,0 +1,86 @@
...
+	X_RATE_LIMIT_REMAINING = "x-rate-limit-remaining"
+	X_RATE_LIMIT_RESET     = "x-rate-limit-reset"
+)
+
+type GovernedTransport struct {
+	base     http.RoundTripper
+	apiMutex *apimutex.ApiMutex
+}
+
+// newRequestThrottleTransport returns RoundTripper which provides throttling according to maxRequests.
```
monde: document public methods and fix typos

PR #596, opened 2021-08-18
okta_app_oauth: mark refresh_token_rotation as computed

bendrucker:
We're seeing a perpetual diff with `okta_app_oauth`, where `refresh_token_rotation` is omitted in the configuration, and is perpetually reset from `"STATIC"` (in state) to `null` (in config).

```
  # okta_app_oauth.app will be updated in-place
  ~ resource "okta_app_oauth" "app" {
        id                         = "0oa44sotcx8nesMuo357"
        name                       = "oidc_client"
      - refresh_token_rotation     = "STATIC" -> null
        # (25 unchanged attributes hidden)
    }
```

I believe marking this value as `Computed` should take care of this:

https://github.com/hashicorp/terraform/issues/21278

Per the [Okta API docs](https://developer.okta.com/docs/reference/api/apps/#refresh-token-object), this is a server-side computed default that differs depending on other fields:

> The default rotation_type value is ROTATE for Single-Page Applications (SPAs). For all other clients, the default is STATIC.


okta/resource_okta_app_oauth.go
```diff
@@ -254,6 +254,7 @@ func resourceAppOAuth() *schema.Resource {
 			"refresh_token_rotation": {
 				Type:             schema.TypeString,
 				Optional:         true,
+				Computed:         true,
```
bogdanprodan-okta, line 271: It's better to set `Default: "STATIC"`, since you can set `refresh_token_rotation` to `"ROTATE"` (however, this is an Early Access Feature)
```
resource "okta_app_oauth" "test" {
  label                  = "test"
  status                 = "ACTIVE"
  type                   = "browser"
  grant_types            = ["authorization_code", "refresh_token"]
  redirect_uris          = ["http://d.com/aaa"]
  response_types         = ["code"]
  hide_ios               = true
  hide_web               = true
  auto_submit_toolbar    = false
  refresh_token_rotation = "ROTATE"
  refresh_token_leeway   = 30
}
```

bogdanprodan-okta (changes requested):

okta/resource_okta_app_oauth.go
bendrucker: Is it? It's certainly an option, but the Okta docs say it's only the default for certain app types, while SPAs (browser?) default to `ROTATE`. Terraform's `Default` field cannot accommodate any sort of conditional logic. 

Happy to change this but it seems like this would be a potential breaking change: users with apps where the server assigned `ROTATE` would see those resources changed to `STATIC`.

bogdanprodan-okta: You can try to create these two test apps via API (one with ROTATE, and one with STATIC)
```
{
    "name": "oidc_client",
    "label": "test",
    "status": "ACTIVE",
    "visibility": {
        "hide": {
            "iOS": true,
            "web": true
        },
        "appLinks": {
            "oidc_client_link": true
        }
    },
    "signOnMode": "OPENID_CONNECT",
    "settings": {
        "app": {},
        "oauthClient": {
            "redirect_uris": [
                "http://d.com/aaa"
            ],
            "response_types": [
                "code"
            ],
            "grant_types": [
                "authorization_code",
                "refresh_token"
            ],
            "application_type": "browser",
            "consent_method": "TRUSTED",
            "issuer_mode": "ORG_URL",
            "refresh_token": {
                "rotation_type": "STATIC"
            },
            "idp_initiated_login": {
                "mode": "DISABLED",
                "default_scope": []
            },
            "wildcard_redirect": "DISABLED"
        }
    }
}
```
And
```
{
    "name": "oidc_client",
    "label": "test",
    "status": "ACTIVE",
    "visibility": {
        "hide": {
            "iOS": true,
            "web": true
        },
        "appLinks": {
            "oidc_client_link": true
        }
    },
    "signOnMode": "OPENID_CONNECT",
    "settings": {
        "app": {},
        "oauthClient": {
            "redirect_uris": [
                "http://d.com/aaa"
            ],
            "response_types": [
                "code"
            ],
            "grant_types": [
                "authorization_code",
                "refresh_token"
            ],
            "application_type": "browser",
            "consent_method": "TRUSTED",
            "issuer_mode": "ORG_URL",
            "refresh_token": {
                "rotation_type": "ROTATE",
                "leeway": 30
            },
            "idp_initiated_login": {
                "mode": "DISABLED",
                "default_scope": []
            },
            "wildcard_redirect": "DISABLED"
        }
    }
}
```

bendrucker: Thanks for those, the test case I have in mind is this:

```json
{
    "name": "oidc_client",
    "label": "test",
    "status": "ACTIVE",
    "visibility": {
        "hide": {
            "iOS": true,
            "web": true
        },
        "appLinks": {
            "oidc_client_link": true
        }
    },
    "signOnMode": "OPENID_CONNECT",
    "settings": {
        "app": {},
        "oauthClient": {
            "redirect_uris": [
                "http://d.com/aaa"
            ],
            "response_types": [
                "code"
            ],
            "grant_types": [
                "authorization_code",
                "refresh_token"
            ],
            "application_type": "browser",
            "consent_method": "TRUSTED",
            "issuer_mode": "ORG_URL",
            "idp_initiated_login": {
                "mode": "DISABLED",
                "default_scope": []
            },
            "wildcard_redirect": "DISABLED"
        }
    }
}
```

Essentially, your first case but with the `refresh_token` unset. The response I get, abbreviated for clarity:

```json
{
  "settings": {
    "oauthClient": {
      "application_type": "browser",
      "client_uri": null,
      "consent_method": "TRUSTED",
      "grant_types": [
        "authorization_code",
        "refresh_token"
      ],
      "idp_initiated_login": {
        "default_scope": [],
        "mode": "DISABLED"
      },
      "issuer_mode": "ORG_URL",
      "logo_uri": null,
      "redirect_uris": [
        "http://d.com/aaa"
      ],
      "refresh_token": {
        "leeway": 30,
        "rotation_type": "ROTATE"
      },
      "response_types": [
        "code"
      ]
    }
  }
}
```

Note that `settings.refresh_token.rotation_type` is `"ROTATE"`.

Given the following configuration:

```tf
resource "okta_app_oauth" "app" {
  type = "browser"
}
```

Currently, that should result in `refresh_token` being set to `"ROTATE"` in state, since the server returns a default. There will be a perpetual diff given that the user-specified value is `null`. If the `Default` were changed, the next apply would change the setting to `"STATIC"` for the live app. 

The change I've proposed here should eliminate the perpetual diff without causing changes to existing resources. Let me know if I'm missing something though.

bendrucker: Hi, just want to re-raise this, still seems like marketing this key as computed is the appropriate fix

bogdanprodan-okta (approved):

PR #615, opened 2021-08-27
Add pagination when getting user groups

BrentSouza:
We have users with a lot of groups assigned. More than the 200 default limit. When getting groups, we're seeing transient errors due to whether or not a particular group we're interested in is present in that first set of 200. This will use the okta client's built in pagination support to keep fetching groups until there are no more pages.

### DIFF
--- a/okta/user.go
+++ b/okta/user.go
@@ -290,14 +290,29 @@ func setAdminRoles(ctx context.Context, d *schema.ResourceData, m interface{}) e
 
 // set all groups currently attached to the user
 func setAllGroups(ctx context.Context, d *schema.ResourceData, c *okta.Client) error {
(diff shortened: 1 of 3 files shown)


bogdanprodan-okta (approved):

BrentSouza:
I felt like including setup for an acceptance test that provisioned 200+ groups might be a bit much so I went with a unit test approach to verify the functionality. If we need to show this working via acceptance tests I can adjust.

bogdanprodan-okta:
> I felt like including setup for an acceptance test that provisioned 200+ groups might be a bit much so I went with a unit test approach to verify the functionality. If we need to show this working via acceptance tests I can adjust.

That's fine as it is :)

PR #643, opened 2021-09-13
[WIP] Add Device Authorization grant type to okta_oauth_app resource

ConnorWidtfeldt:
First PR here, so I may have missed a step or two in the contributing doc.

Native OAuth apps can use the "Device Authorization" grant if that early access feature is enabled. In our case it is, and we have to add it manually using the Okta console, only for it to be removed if our automation is run. It is an early access feature, but it appears to be nearing release and would be nice to include.

### DIFF
--- a/okta/resource_okta_app_oauth.go
+++ b/okta/resource_okta_app_oauth.go
@@ -23,11 +23,12 @@ type (
 
 // I wish the SDK provided these
 const (
-	authorizationCode string = "authorization_code"
-	implicit          string = "implicit"
-	password          string = "password"
-	refreshToken      string = "refresh_token"
-	clientCredentials string = "client_credentials"
+	authorizationCode   string = "authorization_code"
+	implicit            string = "implicit"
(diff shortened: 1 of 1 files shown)


ConnorWidtfeldt:
Didn't mean to move this from a draft state, but don't see a way to go back.

bogdanprodan-okta:
> Didn't mean to move this from a draft state, but don't see a way to go back.

I'll just label this PR with `wip`, so you can remove it once it's done :)

bogdanprodan-okta:
I've created another PR and this feature will be available in the new release.

ConnorWidtfeldt:
Oh sweet! Sorry for the lack of activity, I didn't know what steps still remained to move this forward. I have been testing this locally for the past month and have had no problems so far.

bogdanprodan-okta:
This is the [PR](https://github.com/okta/terraform-provider-okta/pull/694) where I've added this value. It will be released next week.

PR #697, opened 2021-10-08
Added new `okta_email_sender` and `okta_email_sender_verification` resources

### LINKED ISSUE #469: Email & SMS Sender
Is there actually a resource present to configure email and SMS sender via Terraform?
If not please add it.

image

bogdanprodan-okta:
Fix #469


monde (approved):

rtlong:
This is great, if it does what I think it does. But I'm not seeing any documentation for this API at `/api/v1/org/email/sender` (assuming on the Org page but even with Google I'm just not getting much with that string). @bogdanprodan-okta if am I correct in thinking you work at Okta, could you pass on my wish that we see this documented?

PR #699, opened 2021-10-08
fix for mutex key normalization

### LINKED ISSUE #696: max_api_capacity not working properly
### Community Note

### Terraform Version

Terraform v1.0.6
on darwin_amd64
\+ provider registry.terraform.io/okta/okta v3.14.0

### Affected Resource(s)

- all?

```hcl
provider "okta" {
  org_name         = "..."
  base_url         = "okta.com"
  api_token        = "... redacted ..."
  max_api_capacity = 80
}
```

### Debug Output

https://gist.github.com/phi1ipp/c8b01eab5f32ec6bfe96afc418aa4ed1

### Expected Behavior

Refresh state operation should be executed as there is a lot of unused capacity. 
### Actual Behavior
Terraform pauses with a message to sleep for XX seconds to preserve the limit. When I saw a message `sleeping for 60 seconds until rate limit reset ("app-id":98/1800` and then inspected all API calls above it in the log, I was not able to find where remaining `X-Rate-Limit-Remaining` would be 98 and `X-Rate-Limit-Limit` would be 1800.

Other suspicious factor is, that `/api/v1/apps` is not 1800 for sure, even though we have increased limit for this endpoint

### Steps to Reproduce

1. `terraform apply`

phi1ipp:
fixed normalizeKey for `/api/v1/apps/xxxxxxx/[users|groups|sso]` cases

### DIFF
--- a/okta/internal/apimutex/apimutex.go
+++ b/okta/internal/apimutex/apimutex.go
@@ -124,10 +124,10 @@ func (m *APIMutex) Status(method, endPoint string) *APIStatus {
 func (m *APIMutex) normalizeKey(method, endPoint string) string {
 	var result string
 	switch {
-	case endPoint == "/api/v1/apps":
-		result = APPS_KEY
-	case strings.HasPrefix(endPoint, "/api/v1/apps/"):
+	case regexp.MustCompile("/api/v1/apps/[^/]+$").MatchString(endPoint):
 		result = APPID_KEY
+	case strings.Contains(endPoint, "/api/v1/apps"):
+		result = APPS_KEY
 	case endPoint == "/api/v1/certificateAuthorities":
 		result = CAS_KEY
 	case endPoint == "/oauth2/v1/clients":
--- a/okta/internal/apimutex/apimutex.go
+++ b/okta/internal/apimutex/apimutex.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"regexp"
 )
 
 const (


okta/internal/apimutex/apimutex.go
```diff
@@ -124,10 +124,10 @@ func (m *APIMutex) Status(method, endPoint string) *APIStatus {
 func (m *APIMutex) normalizeKey(method, endPoint string) string {
 	var result string
 	switch {
-	case endPoint == "/api/v1/apps":
-		result = APPS_KEY
-	case strings.HasPrefix(endPoint, "/api/v1/apps/"):
+	case regexp.MustCompile("/api/v1/apps/[^/]+$").MatchString(endPoint):
```
monde, line 128: @phi1ipp some things come to mind:

- Can you show there is a bug here with a unit test?
- `regexp.MustCompile` will cause a panic, I don't think we want to the provider to be that brittle https://pkg.go.dev/regexp#MustCompile
- `strings.HasPrefix` doesn't compile a regex, just does a 1-1 character comparison https://github.com/golang/go/blob/master/src/strings/strings.go#L443-L446
- must compile is computational more complex than hasprefix https://github.com/golang/go/blob/master/src/regexp/regexp.go#L169-L218

okta/internal/apimutex/apimutex.go
```diff
@@ -124,10 +124,10 @@ func (m *APIMutex) Status(method, endPoint string) *APIStatus {
...
 		result = APPID_KEY
+	case strings.Contains(endPoint, "/api/v1/apps"):
```
monde, line 130: I'm not seeing a logical difference in the PR doing a code review. Can you show there is a bug with a unit test?

Also, `strings.Contains` will look for a match in the entire string whereas `endPoint == "/api/v1/apps"` is explicit in its logic.

monde:
I think I see the logical flaw re-reading https://developer.okta.com/docs/reference/rl-global-mgmt/ and @phi1ipp's regexp `regexp.MustCompile("/api/v1/apps/[^/]+$")`, I'll work on a PR.

monde:
@phi1ipp @bogdanprodan-okta , also, this is an opportunity for me to also double check another optimization I've been thinking about for rate limiting.

okta/internal/apimutex/apimutex.go
phi1ipp: @monde , the problem is with parallelism, so it's not easily reproducible

Let me explain what I discover while running the code with a lot of debugging statements injected here and there.

When I saw that it's waiting for reset b/c `95/1800` left in mutex for `app_id` I started looking which API call could write such values into apiMutex. Of course there was none for that combination, but 95 would be a limit remaining for `/api/v1/xxxx/groups?limit=xx` call, where `1800` is coming from `/api/v1/apps/xxxxxxx` 

So I started checking this thing and that's fix suggested takes care of that, to segregate `/api/v1/apps/{appId}` calls for which limit by default is 600, from `/api/v1/apps/{appId}/groups` which limit is 100 (see https://developer.okta.com/docs/reference/rl-global-mgmt/)

monde: There is a mutex on the struct that holds the path accounting / rate limits. I think the bug is that it was not taking into account paths for the `/api/v1/apps/xxxxxxx/[users|groups|sso]` cases. This made it look like a parallelism bug but I don't think it is that. I added more information to the sleeping log line in my PR #700

phi1ipp: @monde ,
I know, that there is a mutex. But it only acts during `Update`, so concurrent updates are not allowed. But other goroutine can still read apiMutex, when update is happening. And you can't guarantee, that it won't read it in the middle of the update.

I think my observations proved it, as how else would it get status 95/1800. It should be then 95/100, or 1795/1800.

Just thinking outload

PR #709, opened 2021-10-13
fixing normalizeKeys function

phi1ipp:
fixes to address `/api/v1/[apps|groups]/xxxxxxxx/..` endpoints

### DIFF
--- a/okta/internal/apimutex/apimutex.go
+++ b/okta/internal/apimutex/apimutex.go
@@ -128,10 +128,10 @@ var reGroupId = regexp.MustCompile("/api/v1/groups/[^/]+$")
 func (m *APIMutex) normalizeKey(method, endPoint string) string {
 	var result string
 	switch {
-	case endPoint == "/api/v1/apps":
-		result = APPS_KEY
 	case reAppId.MatchString(endPoint):
 		result = APPID_KEY
+	case strings.Contains(endPoint, "/api/v1/apps"):
+		result = APPS_KEY
 	case endPoint == "/api/v1/certificateAuthorities":
 		result = CAS_KEY
 	case endPoint == "/oauth2/v1/clients":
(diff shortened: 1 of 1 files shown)


monde:
@phi1ipp this change breaks the `TestNormalizeKey` test. Just want to double check your understanding of the API docs given your real world experience.  Here is the broken test:
```
go test -test.v -run "TestNormalizeKey" ./okta/internal/apimutex
=== RUN   TestNormalizeKey
    apimutex_test.go:209: got "apps", expected "other" for method: "GET", endPoint: "/api/v1/apps/XXX/users"
--- FAIL: TestNormalizeKey (0.00s)
FAIL
FAIL    github.com/okta/terraform-provider-okta/okta/internal/apimutex  0.335s
FAIL
```
The API docs https://developer.okta.com/docs/reference/rl-global-mgmt/ list the following:
```
Create or list applications:
/api/v1/apps except /api/v1/apps/${id}

Get, update, or delete an application by ID:
/api/v1/apps/${id} only

Most other API actions:
/api/v1
```
It seems to me `/api/v1/apps/XXX/users` should be in `other` according to the docs but you think you are seeing that it is accounted for as a `/api/v1/apps/${id}` path?
I'll go look at our internal Okta core code and see what is going on there later today but wanted to show you this first.
I pulled your commit into this branch https://github.com/okta/terraform-provider-okta/tree/phi1ipp_fix_api_limits

phi1ipp:
@monde , well, it does make sense to some degree, what you are saying, but...

If you check rate limits table, then you'll noice that `other` may go with 600 per minute, and what I see in my response to `/api/v1/apps/xxxx/users` returns back 100 as a limit, so mixing both of those into one bucket might not be a good idea

monde:
@phi1ipp @bogdanprodan-okta just a follow up. I did a code review of the Okta API with our lead engineer responsible for it. He gave me some insights for a better implementation of rate limit accounting. For instance an Okta Org can have custom rate limits applied to it outside of the rules published in the documentation. I'll have to work out a better PR here.

bogdanprodan-okta:
These changes are captured in https://github.com/okta/terraform-provider-okta/pull/719

PR #899, opened 2022-01-11
Support Okta Managed certificates

### LINKED ISSUE #882: Allow `okta_domain_verification` to succeed when existing status is "COMPLETED"
There are now 2 domain options when creating a domain in the API with the new EA Lets Encrypt / Okta managed domains. I believe that the `okta_domain` may need to be updated to handle both variations.

Additionally, verification may come outside or via UI. Thus, we should be able to create the verification record and have it succeed since it can't be imported. That way it can still be managed via Terraform in the event that at some point it fails.

### Sample configuration
```
resource "okta_domain" "default" {
  name = "okta.my-custom-domain.com"
}

# DNS records handled externally, validated externally

resource "okta_domain_verification" "default" {
  domain_id = okta_domain.default.id
}
```

### Output
```
╷
│ Error: failed to verify domain after several attempts, current validation status: COMPLETED
│
│   with okta_domain_verification.default,
│   on customizations-domain.tf line 9, in resource "okta_domain_verification" "default":
│    9: resource "okta_domain_verification" "default" {
│
╵
```

The expected output should be that this succeeds because it's already complete.

### Community Note

### Terraform Version

```
Terraform v1.1.2
on windows_amd64
+ provider registry.terraform.io/okta/okta v3.20.2
```

### Affected Resource(s)

- okta_domain_verification

virgofx:
Fixes #882 

Confirmed locally testing creation of new `OKTA_MANAGED` cert and getting the resource to create/manage it. The READ wasn't necessary as originally suspected it was since I guess we still won't import but handle it first time on create which works properly handling the other success case `COMPLETED`.

Backwards compatible with the addition of the `certificate_source_type` parameter by using existing default value of `MANUAL` and setting as optional. Ideally, it might make sense to change this to **required** on next version bump and remove the `MANUAL` as default as it's really an explicit user choice (and this mirrors the Okta API). If that were a preference for next version, I could quickly update.

/cc @bogdanprodan-okta

### DIFF
--- a/website/docs/r/domain.html.markdown
+++ b/website/docs/r/domain.html.markdown
@@ -24,7 +24,11 @@ The following arguments are supported:
 
 - `name` - (Required) Custom Domain name.
 
-- `verify` - (Optional) Indicates whether the domain should be verified. 
+- `certificate_source_type` - (Optional) Certificate source type that indicates whether the certificate is provided by the user or Okta. Accepted values: `MANUAL`, `OKTA_MANAGED`. Default value = `MANUAL`
+
+  ~> **WARNING**: Use of `OKTA_MANAGED` requires a feature flag to be enabled.
+
+- `verify` - (Optional) Indicates whether the domain should be verified.
   - `DEPRECATED`: Please use `okta_domain_verification` resource instead.
 
 ## Attributes Reference
--- a/okta/resource_okta_domain.go
+++ b/okta/resource_okta_domain.go
@@ -25,6 +25,12 @@ func resourceDomain() *schema.Resource {
 				Description: "Custom Domain name",
 				ForceNew:    true,
 			},
+			"certificate_source_type": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Optional. Certificate source type that indicates whether the certificate is provided by the user or Okta. Accepted values: MANUAL, OKTA_MANAGED. Warning: Use of OKTA_MANAGED requires a feature flag to be enabled. Default value = MANUAL",
+				Default:     "MANUAL",
+			},
 			"verify": {
 				Type:        schema.TypeBool,
 				Optional:    true,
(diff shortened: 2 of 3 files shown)


bogdanprodan-okta:
Hi, @virgofx! Really appreciate the PR. Could you also add the new field to the corresponding [documentation](https://github.com/okta/terraform-provider-okta/blob/master/website/docs/r/domain.html.markdown#argument-reference)?

virgofx:
Whups, missed that. Should be good now.

website/docs/r/domain.html.markdown
```diff
@@ -24,7 +24,11 @@ The following arguments are supported:
 
 - `name` - (Required) Custom Domain name.
 
-- `verify` - (Optional) Indicates whether the domain should be verified. 
+- `certificate_source_type` - (Optional) Certificate source type that indicates whether the certificate is provided by the user or Okta. Accepted values: `MANUAL`, `OKTA_MANAGED`. Default value = `MANUAL`
+
+  > Warning: Use of `OKTA_MANAGED` requires a feature flag to be enabled.
```
bogdanprodan-okta:
```
~> **WARNING:** Use of `OKTA_MANAGED` requires a feature flag to be enabled.
```

virgofx: Try a refresh now.

bogdanprodan-okta: you're missing `~` at the beginning

virgofx: Interesting. Haven't seen that Markdown syntax before (The `~>`).  Regardless, updated :)

Originally should be similar to:
image

New version looks like:
image

bogdanprodan-okta: Yeah, that's for Hashicorp's registry website

bogdanprodan-okta (approved):

website/docs/r/domain.html.markdown
virgofx: That makes sense. Good to know. Thanks!

PR #924, opened 2022-01-27
fix okta_group_memberships import

exitcode0:
This is my first PR not just for this repo but for Golang in general so please be gentle :) 

#### Reasoning for Change:
manually crafting okta_group_memberships resources for large groups can be time consuming
the TF import -> strip state only schema -> `terraform state show -no-color <address> |pbcopy` workflow is convenient in these scenarios
supporting import also allows for plans to evaluate to  no changes, this makes it much easier for teammates who might be less TF savvy to review your PRs when importing your current Okta configuration

- [ ]  Acceptance test coverage of new behavior
I've tested this resource manually in a non-prod env of mine and it appears to work as expected, not entirely familiar with how to run the acceptance test against the resource with my changes
- [X] Documentation updates
- [ ] Well-formed Code
Most of the code in this change was lifted from the [syncGroupUsers](https://github.com/okta/terraform-provider-okta/blob/0d5cc7a7cc2bbe2018311de2a291ccf976518a45/okta/resource_okta_group.go#L180-L183) function, hopefully this is up to grade, if not i'm open to guidance

I've tested this resource in a non-prod env of mine and it appears to work as expected
If I need to write additional tests for this then i'm happy to learn how to do that and modify the PR

I haven't


okta/resource_okta_group_memberships.go
```diff
@@ -66,20 +68,16 @@ func resourceGroupMembershipsCreate(ctx context.Context, d *schema.ResourceData,
...
 
 func resourceGroupMembershipsRead(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
-	groupId := d.Get("group_id").(string)
-	users := convertInterfaceToStringSetNullable(d.Get("users"))
-	client := getOktaClientFromMetadata(m)
-	ok, err := checkIfGroupHasUsers(ctx, client, groupId, users)
+	var diags diag.Diagnostics
+
+	logger(m).Info("reading group membership", "id", d.Id())
+	userIDList, err := listGroupUserIDs(ctx, m, d.Id())
```
monde: I'm going to defer to @bogdanprodan-okta , it seems like removing the `checkIfGroupHasUsers(` call changes the behavior even though that is the justification from the PR proposal.

okta/resource_okta_group_memberships.go
```diff
@@ -66,20 +68,16 @@ func resourceGroupMembershipsCreate(ctx context.Context, d *schema.ResourceData,
 }
 
 func resourceGroupMembershipsRead(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
-	groupId := d.Get("group_id").(string)
-	users := convertInterfaceToStringSetNullable(d.Get("users"))
-	client := getOktaClientFromMetadata(m)
-	ok, err := checkIfGroupHasUsers(ctx, client, groupId, users)
+	var diags diag.Diagnostics
```
monde: We can punt having a `diags` variable defined here as it isn't used anywhere in the method other than the return statement. Instead, just return nil as the final statement. `diag.Diagnostics` is a type defined as a slice of `diag.Diagnostic` so `return nil` is equivalent to `return []diag.Diagnostic{}` at runtime.

monde:
Thanks for the PR @exitcode0, that is cool you are putting in some work. I'm going to defer to @bogdanprodan-okta about the behavior.

monde:
@exitcode0 I'll review this PR with @bogdanprodan-okta and follow up from there

monde:
@exitcode0 let me know where you are at on this PR and the feedback given, thanks.

exitcode0:
@monde Apologies for letting this one go stale
I've made some additions to the PR, let me know if there is anything that still requires further modification

okta/resource_okta_group_memberships.go
```diff
@@ -123,9 +130,9 @@ func checkIfGroupHasUsers(ctx context.Context, client *okta.Client, groupId stri
 	if err := suppressErrorOn404(resp, err); err != nil {
 		return false, fmt.Errorf("unable to return membership for group (%s) from API", groupId)
 	}
-	if len(groupUsers) == 0 {
-		return false, nil
-	}
+	// if len(groupUsers) == 0 {
+	// 	return false, nil
+	// }
```
monde: On the commit history I see Bogdan added this as a panic bumper in #795 but I also see how that changed the behavior to #1072 . When I look at the [checkIfGroupHasUsers](https://github.com/okta/terraform-provider-okta/blob/18210979890cbc9899e711ef29ea7071a453a3f5/okta/resource_okta_group_memberships.go#L128-L167) I don't see a codepath where a panic could happen so I think this is OK.

However, just delete the lines.

monde: @exitcode0 I'll approve this PR after that commented code is clipped out. Thanks!

monde (approved):
👍 Thank you @exitcode0

monde:
Released as [v3.28.0](https://github.com/okta/terraform-provider-okta/releases/tag/v3.28.0).

PR #925, opened 2022-01-27
feat: add support for samaccountname

achachw:
Hi there,
A simple contribution to add support for samaccountname as user attribut.
What do you think about ?

### DIFF
--- a/examples/okta_user/all_attributes.tf
+++ b/examples/okta_user/all_attributes.tf
@@ -23,6 +23,7 @@ resource "okta_user" "test" {
   preferred_language = "en-us"
   primary_phone      = "4445556666"
   profile_url        = "http://www.example.com/profile"
+  samaccountname     = "tsmith"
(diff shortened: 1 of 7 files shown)


virgofx:
@achachw I'm not sure the purpose of this. It appears based on [this](https://help.okta.com/en/prod/Content/Topics/Directory/Directory_AD_Field_Mappings.htm) that the samaccountname is derived. If you want it as an attribute on the user, just add a user schema property yourself and you're set. Wouldn't that solve your issue?

achachw:
@virgofx  Thanks for your feedback.
I will need to check how this "samaccountname" property is added and try to understand further this "schema property". 
Basically with this MR i tried to get from datasource this attribut.

achachw:
@virgofx  Finally this attribut is now added as schema property. We will use it.
Thanks and sorry for the noise.

virgofx:
Awesome. I'm glad you have a workaround and thanks for the follow up!

Issue #944, opened 2022-02-03
The okta_groups data source has a hard cap of 300 groups returned


jyoder-deloitte:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

Terraform v1.0.8
on darwin_amd64

provider registry.terraform.io/hashicorp/aws v3.74.0
provider registry.terraform.io/okta/okta v3.20.5

### Affected Resource(s)

- data.okta_groups

### Terraform Configuration Files

```hcl
data "okta_groups" "apps" {
  q    = "app-"
  type = "OKTA_GROUP"
}

output "app_group_count" {
  description = "Okta App Group Count"
  value = length(data.okta_groups.apps.groups)
}
```

### Expected Behavior

A full list of all matching groups should be returned in the `data.okta_group.apps.groups` list

### Actual Behavior

Only the first 300 matching groups are returned in the `data.okta_group.apps.groups` list

### Steps to Reproduce

1. `terraform plan`
2. Examine the `app_group_count` output value

### Important Factoids

Thank you for the prompt fix to #932!  Unfortunately the fix only got me another 100 groups. I still have at least 115 more to go.

### References

- #932
- #938

monde:
Thanks @jyoder-deloitte. I can see what is causing this. The provider is [making a query](https://github.com/okta/terraform-provider-okta/blob/master/okta/data_source_okta_groups.go#L69) to the find groups API endpoint via the okta-sdk-golang package and API's default max results returned is [300 on the find query](https://developer.okta.com/docs/reference/api/groups/#find-groups). We can fix this in a PR.

jyoder-deloitte:
@monde @MikeMondragon-okta @bogdanprodan-okta  Thanks for promptly fixing this, but it looks like it just missed the v3.20.6 release.  Can you create a v3.20.7?

monde:
@jyoder-deloitte I'll cut a v3.20.7 release in the next hour and report back here.

monde:
@jyoder-deloitte I'm double checking some additional items in the provider before my original estimate of a quick release. However, this release will be going out at some time today.

[...]

PR #946, opened 2022-02-04
Set a high limit on the underlying `client.Group.ListGroups` query.

monde:
Addressing #944


virgofx:
Looks good - but putting on devils advocate hat here. What if someone has 100,000 groups. Should we enact pagination here or not even worth it? Again, I think this is a pretty high number so it should be okay, but just throwing it out there ;P

bogdanprodanj:
> Looks good - but putting on devils advocate hat here. What if someone has 100,000 groups. Should we enact pagination here or not even worth it? Again, I think this is a pretty high number so it should be okay, but just throwing it out there ;P

The problem is that when using `q` for group search pagination is not available

bogdanprodan-okta (approved):

PR #948, opened 2022-02-04
CI Job 

ymylei:
This adds a job for master / prs which would:
- Check formatting
- Check Go vetting (and staticcheck)
- Check linting
- Test build
- Run non-acceptance (non-live tests)

Would like to see this expanded on, but some of this basic stuff should be useful to keep the repo in a good state.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,36 @@
+name: CI
+on:
+  push:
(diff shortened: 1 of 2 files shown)


monde:
@ymylei I like this PR, I will discuss it with @bogdanprodan-okta

monde (approved):
👍

Issue #1043, opened 2022-04-04
activate/deactivate default auth server failed due to id not found


peijiinsg:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

Terraform v1.0.0

### Affected Resource(s)

- okta_auth_server_default

### Terraform Configuration Files

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key: https://keybase.io/hashicorp
resource "okta_auth_server_default" "okta_default" {
  name        = "default"
  status      = "INACTIVE"
  description = "Deactivated default auth server"
}
```

### Debug Output

```bash

-----------------------------------------------------: timestamp=2022-04-04T07:30:35.809Z
--
  | 2022-04-04T07:30:35.810Z [DEBUG] provider.terraform-provider-okta_v3.22.1: 2022/04/04 07:30:35 [DEBUG] performing request: method=POST url=https://xxx.oktapreview.com/api/v1/authorizationServers//lifecycle/deactivate
  | 2022-04-04T07:30:35.810Z [INFO]  provider.terraform-provider-okta_v3.22.1: 2022/04/04 07:30:35 [DEBUG] Okta API Request Details:
  | ---[ REQUEST ]---------------------------------------
  | POST /api/v1/authorizationServers//lifecycle/deactivate HTTP/1.1
  | Host: xxx.oktapreview.com
  | User-Agent: okta-sdk-golang/2.9.2 golang/go1.17.7 linux/amd64 okta-terraform/3.22.1
  | Content-Length: 0
  | Accept: application/json
  | Authorization: SSWS ****
  | Accept-Encoding: gzip
```

### Expected Behavior

The default auth server is getting disabled

### Actual Behavior

Request to deactivate the default auth server failed, because the POST request URL is not constructed correctly.

### Steps to Reproduce

1. `terraform apply`

### References

- #0000

monde:
I'll be looking into this and your PR, thanks @peijiinsg

monde:
[...]

PR #1045, opened 2022-04-04
Fix default auth server id when activate/deactivate it

peijiinsg:
This PR is to fix the issue: https://github.com/okta/terraform-provider-okta/issues/1043
When activate/deactivate default auth server, the id is not fetched properly which results an invalid POST url constructed.


monde (approved):
Looks good, I see the logical error if id is set by name `id = d.Get("name").(string)` but later represented by `d.Id()` which could be blank.

peijiinsg:
> Looks good, I see the logical error if id is set by name `id = d.Get("name").(string)` but later represented by `d.Id()` which could be blank.

yeah, I got the blank value when trying to deactivating the auth server, which result the error.

monde:
@peijiinsg  this is released as [v3.23.0](https://github.com/okta/terraform-provider-okta/releases/tag/v3.23.0)

PR #1076, opened 2022-04-25
docs/r/okta_app_saml: Minor updates to documentation for r/okta_app_saml

jphuynh:
I was originally looking to fix the Optional part for `acs_endpoints` but then did an overall pass for the resource docs. Making wording consistent for default values and added some details about required arguments when `preconfigured_app` isn't provided.


monde:
Awesome, thank you @jphuynh, I'll get this in the release I'm about to do.

monde (approved):

PR #1137, opened 2022-05-24
fix data_okta_user omitting admin roles

### LINKED ISSUE #1014: data_source_okta_users omits admin_roles
### Community Note

### Terraform Version
```
Terraform v1.0.11
on darwin_arm64
+ provider registry.terraform.io/okta/okta v3.22.0
```
### Affected Resource(s)

- data_source_okta_users

```hcl
data "okta_users" "BLAH" {
  search {
    name = "profile.login"
    comparison = "eq"
    value = "exact.email.address@domain.com"
  }
}

output "testing" {
  value = data.okta_users.BLAH.users
}

output "testing2" {
  value = [for user in data.okta_users.BLAH.users : user.id if length(user.admin_roles) > 0]
}
```

### Expected Behavior

when the user in question is a super admin, i'd expect `testing2` output to contain the users ID and `testing1` to contain an entry similar to `"admin_roles" = toset(["SUPER_ADMIN"])`

### Actual Behavior

the output `testing2` outputs as empty and `testing1` correctly outputs the user object but the admins roles entry shows as `"admin_roles" = toset([])`

### Steps to Reproduce

1. `terraform refresh`

### References

- #0000

exitcode0:
fixes: #1014 

one oddity that I wasn't able to figure out is that when `skip_groups = true` groups is omitted from the state completely
when `skip_roles = true` admin_roles is still present in state as an empty list. Not sure if the field should be omitted completely if `skip_roles = true`


okta/data_source_okta_user.go
```diff
@@ -112,22 +112,14 @@ func dataSourceUserRead(ctx context.Context, d *schema.ResourceData, m interface
 		return diag.Errorf("failed to set user's properties: %v", err)
 	}
 
-	skip := false
-	if val := d.Get("skip_roles"); val != nil {
-		skip = val.(bool)
-	}
-	if !skip {
+	if d.Get("skip_roles") == false {
```
monde, line 115: Doesn't this need to be done as a type assertion and using OK to avoid a nil exception?

```go
if ok, sr := d.GetOk("skip_roles"); ok && sr.(bool) {
// code
}
```
or possibly adding an extra check on the type assertion.

```go
if ok, sr := d.GetOk("skip_roles"); ok {
    if skipRoles, ok := sr.(bool); ok && skipRoles {
        // code
    }
}
```

okta/data_source_okta_user.go
```diff
@@ -112,22 +112,14 @@ func dataSourceUserRead(ctx context.Context, d *schema.ResourceData, m interface
...
 			return diag.Errorf("failed to set user's admin roles: %v", err)
 		}
 	}
 
-	skip = false
-	if val := d.Get("skip_groups"); val != nil {
-		skip = val.(bool)
-	}
-	if !skip {
+	if d.Get("skip_groups") == false {
```
monde, line 122: same as previous type assertion comment

okta/user.go
```diff
@@ -258,11 +258,11 @@ func listUserOnlyRoles(ctx context.Context, c *okta.Client, userID string) (user
 		return
 	}
 	for _, role := range roles {
-		if role.AssignmentType == userScope && role.Type != "CUSTOM" {
+		if role.Type != "CUSTOM" {
 			userOnlyRoles = append(userOnlyRoles, role)
 		}
 	}
-	return
+	return userOnlyRoles, resp, err
```
monde: The method ` listUserOnlyRoles` has named return types so `return userOnlyRoles, resp, err` is redundant and should only be `return` (personally, I'm not a fan of named return types given it makes bare return states look odd/illogical)

monde (commented):
@exitcode0 I'd like to hear your thoughts on these comments. Can we get an ACC test?

okta/data_source_okta_user.go
exitcode0: because `skip_roles` [in the schema](https://github.com/okta/terraform-provider-okta/blob/master/okta/data_source_okta_user.go#L70) has a default value I thought this might be okay?
the default value of bool false should prevent the value from ever being nil? 

I'm still very new to golang so happy to take guidance where I can get it
if you think I should update these lines, happy to make what ever changes are required 😄

okta/user.go
exitcode0: i've never seen that functionality before, thanks for the info 😄

monde:
@exitcode0 I started writing tests for this PR and rereading #1014 . TL;DR I ended up doing a new new PR #1159

Some things to note:
- `admin_roles` and `group_memberships` will always be present on an `okta_user` data source because it is defined on its schema
- found a bug on data source `okta_users` where it wasn't including `admin_roles` which is what you were seeing in the debug output of your example for #1014
- added an equivalent `include_roles` to `include_groups` on data source `okta_users`

PR #1142, opened 2022-05-30
Allow configuring a custom endpoint

ido50:
The ability to configure a custom URL endpoint for the Okta API is added
to the provider. This is useful for unit testing or running local
caching proxies.

A new configuration option called "custom_endpoint" is added, which by
default takes the value of the "OKTA_CUSTOM_ENDPOINT" environment
variable. When used, and if the provided endpoint is an HTTP endpoint
(rather than an HTTPS endpoint), the HTTPS check is automatically
disabled.

A unit test called `TestCustomEndpoint` is added which also showcases
unit testing of the Okta API without actually contacting Okta.
Information about the new configuration option is also added to the
documentation.

### DIFF
--- a/okta/config.go
+++ b/okta/config.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/hashicorp/go-cleanhttp"
@@ -33,6 +34,7 @@ type (
 	Config struct {
 		orgName          string
 		domain           string
+		httpProxy        string
 		apiToken         string
 		clientID         string
 		privateKey       string
@@ -90,8 +92,17 @@ func (c *Config) loadAndValidate(ctx context.Context) error {
 		httpClient.Transport = transport.NewGovernedTransport(httpClient.Transport, apiMutex, c.logger)
 	}
 
+	var orgUrl string
+	var disableHTTPS bool
+	if c.httpProxy != "" {
+		orgUrl = strings.TrimSuffix(c.httpProxy, "/")
+		disableHTTPS = strings.HasPrefix(orgUrl, "http://")
+	} else {
+		orgUrl = fmt.Sprintf("https://%v.%v", c.orgName, c.domain)
+	}
+
 	setters := []okta.ConfigSetter{
-		okta.WithOrgUrl(fmt.Sprintf("https://%v.%v", c.orgName, c.domain)),
+		okta.WithOrgUrl(orgUrl),
 		okta.WithToken(c.apiToken),
 		okta.WithClientId(c.clientID),
 		okta.WithPrivateKey(c.privateKey),
@@ -105,6 +116,9 @@ func (c *Config) loadAndValidate(ctx context.Context) error {
 	}
 	if c.apiToken == "" {
 		setters = append(setters, okta.WithAuthorizationMode("PrivateKey"))
+	}
+	if disableHTTPS {
+		setters = append(setters, okta.WithTestingDisableHttpsCheck(true))
 	}
 	_, client, err := okta.NewClient(
 		context.Background(),
--- a/okta/provider.go
+++ b/okta/provider.go
@@ -179,6 +179,12 @@ func Provider() *schema.Provider {
 				DefaultFunc: schema.EnvDefaultFunc("OKTA_BASE_URL", "okta.com"),
 				Description: "The Okta url. (Use 'oktapreview.com' for Okta testing)",
 			},
+			"http_proxy": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("OKTA_HTTP_PROXY", ""),
+				Description: "Alternate HTTP proxy of scheme://hostname or scheme://hostname:port format",
+			},
 			"backoff": {
 				Type:        schema.TypeBool,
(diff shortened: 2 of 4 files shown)


okta/config.go
```diff
@@ -90,8 +92,17 @@ func (c *Config) loadAndValidate(ctx context.Context) error {
 		httpClient.Transport = transport.NewGovernedTransport(httpClient.Transport, apiMutex, c.logger)
 	}
 
+	var orgUrl string
+	var disableHTTPS bool
```
monde, line 96: I like where this is going with having a boolean to mark if we are going to be enabling TLS, or not.

okta/provider.go
```diff
@@ -179,6 +179,12 @@ func Provider() *schema.Provider {
 				DefaultFunc: schema.EnvDefaultFunc("OKTA_BASE_URL", "okta.com"),
 				Description: "The Okta url. (Use 'oktapreview.com' for Okta testing)",
 			},
+			"custom_endpoint": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("OKTA_CUSTOM_ENDPOINT", ""),
+				Description: "Custom URL endpoint for unit testing and caching",
+			},
```
monde, line 187: but I'm not seeing the value of having a `OKTA_CUSTOM_ENDPOINT`. The hostname of the org is already set by `OKTA_ORG_NAME` combined with `OKTA_BASE_URL`.

I think it would be more flexible if the env var that triggered all of this off was `OKTA_HTTP_PROTOCOL`.

For example, if `OKTA_HTTP_PROTOCOL=http`, that would be the thing that is used to set up the `disableHTTPS` in `okta/config.go` and the old code `okta.WithOrgUrl(fmt.Sprintf("https://%v.%v", c.orgName, c.domain)),` would take `disableHTTPS` into consideration.

monde:
@ido50 let me know what you think my suggestion of `OKTA_HTTP_PROTOCOL` . We need you to sign our CLA https://developer.okta.com/cla/ in order to take this PR.

okta/provider.go
ido50: Hi @monde . The `disableHTTPS` boolean isn't the main target of this PR but merely a side effect. I want to have the provider make its requests to a completely custom endpoint that _isn't_ the actual Okta API. Specifically, an endpoint of a proxy that caches responses so I don't hit Okta rate limits. This proxy can be anywhere. The current code does not allow this because it builds the URL from the organization and domain with a hardcoded dot between them, as well as hardcode an "https://" scheme. If the proxy, for example, is local, it can easily be HTTP-only on localhost.

ido50:
Signed and sent CLA. Thanks.

okta/provider.go
monde: Thanks for the further clarification @ido50 . This behavior is a proxy and I'd like to see it named that way here. In the spirit of [httpproxy](https://pkg.go.dev/golang.org/x/net/http/httpproxy) in golang's `net/http` it should have similar environment variables. As all of the env variables that the provider is responsible for are prefixed with `OKTA_` I think a better env variable would be `OKTA_HTTP_PROXY`. We could use it for both http and https values. okta/config.go would have a variable named `httpProxy` that is a string. 

```go
var orgUrl string
var disableHTTPS bool
	
if c.customEndpoint != "" {
  orgUrl = strings.TrimSuffix(c.customEndpoint, "/")
  disableHTTPS = strings.HasPrefix(orgUrl, "http://")
} else {
  orgUrl = fmt.Sprintf("https://%v.%v", c.orgName, c.domain)
}
```

I don't know the historical reason the provider uses two env vars (`OKTA_ORG_NAME` +  `OKTA_BASE_URL`) to make a hostname. FWIW I'd like to change that in the next major version of the provider.

okta/config.go
```diff
@@ -33,6 +34,7 @@ type (
 	Config struct {
 		orgName          string
 		domain           string
+		customEndpoint   string
```
monde: Asking for name `customEndpoint` be named `httpProxy` per the main discussion on the PR.

okta/provider.go
monde: per main discussion the PR
```go
"http_proxy": {
  Optional:    true,
  DefaultFunc: schema.EnvDefaultFunc("OKTA_HTTP_PROXY", ""),
  Description: "Alternate HTTP proxy of scheme://hostname or scheme://hostname:port format",
},

ido50: Done

okta/config.go
ido50: Done

ido50: Also updated the commit message to reflect these changes.

monde (approved):
Excellent idea @ido50

monde:
@ido50 this will be released in 3.29.0 either tomorrow or Friday.

gilcoh33:
> @ido50 this will be released in 3.29.0 either tomorrow or Friday.

Hi @monde - any ETA for the 3.29.0 release?

monde:
hi @gilcoh33 , v3.29.0 will be released either today or tomorrow. I'm trying to get #1149 and #1137 addressed in the release.

monde:
Released https://github.com/okta/terraform-provider-okta/releases/tag/v3.29.0

PR #1171, opened 2022-06-21
Ordering of redirect_uris in resource okta_app_oauth

monde:
In resource `okta_app_oauth` the property`redirect_uris` is a list, not a set, and needs to maintain order.

### DIFF
--- a/okta/resource_okta_app_oauth.go
+++ b/okta/resource_okta_app_oauth.go
@@ -208,7 +208,7 @@ func resourceAppOAuth() *schema.Resource {
 				Description: "List of scopes to use for the request",
 			},
 			"redirect_uris": {
-				Type:        schema.TypeSet,
+				Type:        schema.TypeList,
 				Elem:        &schema.Schema{Type: schema.TypeString},
 				Optional:    true,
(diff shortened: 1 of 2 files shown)


vijetmahabaleshwar-okta (approved):

okta/resource_okta_app_oauth.go
```diff
@@ -734,7 +734,7 @@ func buildAppOAuth(d *schema.ResourceData) *okta.OpenIdConnectApplication {
 			InitiateLoginUri:       d.Get("login_uri").(string),
 			LogoUri:                d.Get("logo_uri").(string),
 			PolicyUri:              d.Get("policy_uri").(string),
-			RedirectUris:           convertInterfaceToStringSetNullable(d.Get("redirect_uris")),
```
garmishr, line 737: Shouldn't we have some check for uniqueness of redirect_uris? Thoughts?

monde: @garmishr I feel like now that the ordering is not modified (is maintained correctly) that any further inspection isn't necessary. If the Okta API doesn't like the input parameters then it will error out and the operator can then go and make corrections to their config.

PR #1189, opened 2022-07-06
fix: use correct api to add an app to an enrollment policy

felixcolaci:
The AddAppToEnrollmentPolicy method used an incorrect API resulting in the operation to fail on my OIE tenant. I switched out the mappings api in favor of the policy subresource of the apps API.


monde:
Bringing these changes into #1191 as our CI will only run on okta/terraform-provider-okta branches.

Issue #1101, opened 2022-05-12
Failed to sync groups and users for OAuth application... context deadline exceeded


hueydn:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

terraform -v
Terraform v1.1.9
on linux_amd64

### Affected Resource(s)

- okta_app_oauth

### Terraform Configuration Files

```hcl
#OIDC Client for Profile Management App
resource "okta_app_oauth" "ProfileMGMT_WEBAPP" {
  label                       = "Profile Management"
  status                      = "ACTIVE"
  type                        = "browser"
  token_endpoint_auth_method  = "none"
  auto_key_rotation           = true
  redirect_uris               = var.pfmgmt_redirect_uris
  post_logout_redirect_uris   = var.pfmgmt_post_logout_redirect_uris
  response_types              = ["code"]
  grant_types                 = ["authorization_code"]
  consent_method              = "REQUIRED"
  issuer_mode                 = "CUSTOM_URL"
  hide_ios                    = true
  hide_web                    = true
  login_mode                  = "DISABLED"

  lifecycle {
     ignore_changes = [groups]
  }
}
```

### Debug Output

https://gist.github.com/huy-keydata/6bf7a15da1888a42c9af9458b7115412

### Expected Behavior

Terraform should have completed successfully, like in other environments where the same terraform code is run

### Actual Behavior

Terraform fails with error in stout provided

### Steps to Reproduce

We have several okta tenants to support our client and use terraform to keep our tenants functional configuration in sync. However we recently went live and currently have over 275k users in our main tenant where this is failing. 

During the "terraform plan" step we get the error you see in the logs provided.

### References

- #0000

monde:
@huy-keydata are you sure you setting up your resource correctly? That error occurs when `implicit_assignment` is not present.

Here are the docs:
https://registry.terraform.io/providers/okta/okta/latest/docs/resources/app_oauth#argument-reference

This the code
https://github.com/okta/terraform-provider-okta/blob/master/okta/resource_okta_app_oauth.go#L513-L519

hueydn:
[...]

PR #1207, opened 2022-07-18
Adds customizable Timeouts to resources/data that rely on syncing users and groups to avoid context.DeadlineExceeded

emanor-okta:
[WIP] Adds customizable Timeouts to resources/data that rely on syncing users and groups (https://www.terraform.io/plugin/sdkv2/resources/retries-and-customizable-timeouts)

Changes the default timeouts from 20min to 1 year (no timeout) for application resources that can sync user/group memberships. This should resolve issues with running into `context.DeadlineExceeded` during the plan/apply phase for Orgs/Apps with large user assignments that don't use federation broker mode.

Addresses issue [#1101](https://github.com/okta/terraform-provider-okta/issues/1101)


monde:
Thanks for the PR @emanor-okta . I'll be able to look at this closer later in the week.

monde:
@emanor-okta disabling timeouts on all the C/R/U's of CRUD doesn't feel right to me. If that was the proper behavior then the TF SDK  would ship with that as its default. Instead, this PR is a substantial behavior change and flips the onus on to the operator to set [`timeouts`](https://www.terraform.io/plugin/sdkv2/resources/retries-and-customizable-timeouts) in their config; perhaps this would be considered a breaking change to some. Also touching 30 files is not DRY at all.

emanor-okta:
@monde - It would be possible to also make the timeout value be 20 minutes `schema.DefaultTimeout(20 * time.Minute)`.
This way it matches the current timeout and the config option would still be available for these resources so an operator could modify it if they ran into issues with timeouts.

emanor-okta:
@monde - I removed the Timeout block from the data_source config files and website files.
For the resource files that make use of the sync user/groups the read/create/delete default timeout has been set to 1 hour which should be enough for most use cases and can always be modified via .tf

monde:
I wrote the ACC tests for this https://github.com/okta/terraform-provider-okta/pull/1267

monde (approved):
:shipit:

PR #1212, opened 2022-07-20
[WIP] Adds ability to flip individual authenticator methods on/off for Authenticator type phone_number

emanor-okta:
[WIP] - OIE
Two Authenticators (phone_number / okta_verify) allow individual methods to be enabled / disabled independent of the status of the authenticator. Currently this is exposed via the [_links](https://developer.okta.com/docs/reference/api/authenticators-admin/#error-response-example) property and is not documented in the Okta API, or any Management SDK.

This is to explore if the Terraform Provider should provide this capability which currently can only be controlled via the UI.
This [WIP] enables this feature for the okta_authenticator resource (CRUD), as well as updates the website with instructions for the phone_number Authenticator type.

If this feature should move forward tests and the data_source additions will be added along with Okta Verifier changes.

### DIFF
--- a/okta/authenticator_methods.go
+++ b/okta/authenticator_methods.go
@@ -1,0 +1,68 @@
+package okta
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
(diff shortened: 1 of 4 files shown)


monde:
Thanks @emanor-okta , I'll take a look.

PR #1215, opened 2022-07-21
export the app embed url for saml apps

felixcolaci:
Hi.
When trying to use the embed url of my saml application in the terraform context I noticed that it hasn't been exported yet.

I suggest to export it as an additional attribute so users are able to use that in their automations. This comes in handy fore example if you want to set this url as a configuration value in another application that is orchestrated with the same terraform playbook.

The goal is that it can be used like so:

```hcl
output "embed_url" {
  value = okta_app_saml.secure_app.embed_url
}
```

Let me know what you think! Cheers.


monde (approved):
Looks good to me @felixcolaci , thank you

monde:

PR #1222, opened 2022-07-27
feat: add `access_token` to the provider config

ericnorris:
This PR adds the ability to configure the provider with an access token directly, rather than with an API key or a private key.

Note: we (Etsy) have been chatting with you (Okta) directly about authentication changes we may want to submit as a PR - this change is related, but different from what we eventually want. This is more of a short to medium term fix for us while we figure out the applicability of our other desired changes with your team. Thanks!


monde:
Thanks @ericnorris . I'd like to see some links to the API documentation at developer.okta.com about the authentication mechanism this is making use of. They would be listed in the docs published to the TF registry that are listed in https://github.com/okta/terraform-provider-okta/blob/master/website/docs/index.html.markdown and also in this PR description.

ericnorris:
Hey @monde! I've pushed a change adding it to the documentation.

>  I'd like to see some links to the API documentation at developer.okta.com about the authentication mechanism this is making use of.

This is using the existing OAuth 2.0 client authentication mechanism that is linked in the docs, so I added a bit about when a caller would be using `access_token` instead of `private_key`, etc. Feel free to tweak the wording if you have suggestions!

monde (approved):
Looks good to me @ericnorris

monde:

Issue #1024, opened 2022-03-16
Error: failed to create group: the API returned an error: The client_assertion JWT kid is invalid.


michaelmisshore:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

Terraform v1.1.7
on darwin_amd64

### Affected Resource(s)

- okta_provider

### Terraform Configuration Files

An Okta Provider configured to use client_id, scopes, and private key followed by any resource creation

### Expected Behavior

A successful API call

### Actual Behavior

Error: failed to create group: the API returned an error: The client_assertion JWT kid is invalid.

### Steps to Reproduce

1. Add an Okta application with multiple JWKs via console
2. Configure an OktaProvider (client_id, scopes, and private key) to use that application with one of the JWKs
3. Attempt to use an resource

### References

https://github.com/okta/okta-sdk-python/issues/289

- #0000

monde:
Thank you @michaelmisshore I will look into this.

PR #1223, opened 2022-07-28
Add `privateKeyId` support.

powellchristoph:
Currently, if you have an application configured with Public key / Private key Client authentication and you have multiple keys configured, the provider will throw the following error:

`Error: failed <insert action>: the API returned an error: The client_assertion JWT kid is invalid.`

This PR allows the `KID` to be set to use the correct key.

### DIFF
--- a/okta/config.go
+++ b/okta/config.go
@@ -38,6 +38,7 @@ type (
 		apiToken         string
 		clientID         string
 		privateKey       string
+		privateKeyId     string
 		scopes           []string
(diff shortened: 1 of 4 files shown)


powellchristoph:
Looks like the same [issue](https://github.com/okta/terraform-provider-okta/issues/1024).

monde (approved):
Awesome, thank you @powellchristoph

monde:

PR #1229, opened 2022-08-01
Fix typo (OKTA_API_PRIVATE_KEY_IE) in provider_test

lukas-hetzenecker:
When looking through the changes of release v3.32.0, we spotted what looks like an obvious typo in provider_test.go.
Introducing this PR to fix it, as such a bug likely becomes a major headache in future.


monde (approved):
Thank you @lukas-hetzenecker

PR #1233, opened 2022-08-01
fix: specify client ID when using private key

ericnorris:
Apologies, it looks like I neglected to add `WithClientID` when using a private key.


monde:
Thanks @ericnorris , our PM asked in Monday morning triage to be sure to have unit tests around TF config input going forward. I'll start a new PR and cherry pick your fix into that and also write the config unit tests.

ericnorris:
Sounds good! Let me know if you need anything from me.

monde:
brought this in to #1234

PR #1237, opened 2022-08-04
wip: configure provider with pre-signed jwt

ericnorris:
This is a proof-of-concept PR following https://github.com/okta/okta-sdk-golang/pull/319 that demonstrates the ability to configure the provider with a pre-signed JWT.  I've left some context in that PR that I will duplicate here:

> As a reminder for context: we'd like to explore using the Okta Terraform provider in an environment where the private key is opaque to us, that is, we have a mechanism for signing JWTs with some private key, but we don't know the actual value of the private key, and so we cannot configure the provider (nor the SDK) with the private key directly.

I decided to split up the work into two PRs to show the most obvious way that this would be implemented, but I am open to changes that keep this localized to the provider if that makes more sense to you all.

As it is now, the changes are pretty superficial - the provider simply passes the JWT through to the SDK, which then uses it to exchange for an access token.

---

Per a conversation we had yesterday, I've attached a rough example of how we (Etsy) plan on calling the provider with a `jwt` argument.

```terraform
provider "google" {}

data "google_service_account_jwt" "okta" {
  target_service_account = "foo@example-gcp-project.iam.gserviceaccount.com"

  payload = jsonencode({
    "aud": "https://foo.okta.com/oauth2/v1/token",
    "iss": "client-id",
    "sub": "client-id",
    "exp": 1234567890
  })
}

provider "okta" {
  jwt = data.google_service_account_jwt.okta.jwt
}
```


monde:
Seems like a good idea to me, my notes from the SDK PR https://github.com/okta/okta-sdk-golang/pull/319#issuecomment-1208336741 "As the API supports this, the SDK should as well. We need to write some tests around this PR and as well as explanation and usage documentation in the README."

monde:
not stale

PR #1239, opened 2022-08-05
Fix data_source_okta_network_zone next page check (limited to 20 zones)

natmariam:
**Issue**: 
- if your Okta instance has more than **20** network zones (`okta_network_zone`), data sourcing will only get the first 20. 
- `data_source_okta_network_zone.go` actually attempts to get more zones if _next page_ (`resp.HasNextPage()`) is available, but actually does not check the set of array returned from next page (`moreZones`) but still contains to access the first page array (`zones`). 

**Solution**:
- Access and compare the data source `name` value from `moreZones` array and not from `zones` array.

### DIFF
--- a/okta/data_source_okta_network_zone.go
+++ b/okta/data_source_okta_network_zone.go
@@ -120,8 +120,8 @@ func findNetworkZoneByName(ctx context.Context, m interface{}, name string) (*ok
 				return nil, err
 			}
 			for i := range moreZones {
-				if zones[i].Name == name {
-					return zones[i], nil
+				if moreZones[i].Name == name {
+					return moreZones[i], nil
 				}
 			}
 		} else {


monde:
@natmariam I like this PR, I'll get it into the next release. I need to write an acceptance test go along with it. I've gotten stung a couple of times recently taking PRs that seem obvious fixes that don't have acceptance covering the code change.

natmariam:
Thank you for the reply and for considering it for next release @monde. 

Put together a simple pay to mock the network zone data and duplicate the bug here: [https://github.com/natmariam/terraform-provider-okta-pr-1239](https://github.com/natmariam/terraform-provider-okta-pr-1239)

monde (approved):
thank you @natmariam

Issue #1279, opened 2022-09-01
client_secret is missing from okta_app_oauth data source


rickardp:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

```
$terraform -v
Terraform v1.2.2
on linux_arm64
```

### Affected Resource(s)

- `okta_app_oauth` data source

### Expected Behavior

When defining an OAuth app, I can receive the client secret trough the `client_secret` attribute.
When referencing an OAuth app defined elsewhere, I can do the same by using the data source with the same name.

### Actual Behavior

The `client_secret` attribute is only available on the resource, not the data source, hence it is difficult to reuse an Okta OAuth app across different IaC projects.

I can get the client secret from the UI and it looks like there is an API for it, so it should be possible to implement.

I am not sure if this is a bug or a feature request, but it is actually the first time I've come across a data source in terraform that does not expose the same set of features as the resource, so I believe this goes against the intent of this data source.

rickardp:
I added a draft PR that adds this fields and make my Terraform IaC work the way I expected it to w.r.t. client_secret.

monde:
@rickardp thanks for the report and the PR illustrating desired behavior.

rickardp:
Any thought on how to get this fixed? I'm happy to help out, continuing on that PR with some guidance on how to best proceed

monde:
@dkulchinsky fixed it in #1285 . This will go out in the next release. I'm going to try and get one out tomorrow.

rickardp:
Nice! Just a heads up, I tried this and got an empty client secret back. I had to specifically request the secrets to get it to work in #1280, though I must admit that this is my first shot at calling the Okta API so I might miss something obvious.

A simple curl to `/api/v1/apps/(appid)` does not reveal a client_secret in credentials/oauthClient, I see only the following keys: "autoKeyRotation", "client_id", "token_endpoint_auth_method"

dkulchinsky:
[...]

PR #1280, opened 2022-09-01
Added support for retrieving client secret from okta_app_oauth data source

rickardp:
Fixes the problem for the specific case reported in #1279.

This works in the case that I have tested, but I am not sure if it is a good idea to fetch the secrets using this call (maybe it causes problems with token scopes) in which case maybe there is a need to opt in/out of this.

Draft PR because it currently serves as discussion material and I do not expect it to be merged in the current state. I can spend some time fixing it up based on review feedback if this is the correct path for solving this issue.

### DIFF
--- a/okta/data_source_okta_app_oauth.go
+++ b/okta/data_source_okta_app_oauth.go
@@ -216,7 +216,7 @@ func dataSourceAppOauthRead(ctx context.Context, d *schema.ResourceData, m inter
 		}
 		// There can be only two client secrets. Choose the latest created one that is active
 		if len(secretList) > 0 {
-			if secretList[0].Status == "ACTIVE" && secretList[1].Status == "ACTIVE" {
+			if len(secretList) > 1 && secretList[0].Status == "ACTIVE" && secretList[1].Status == "ACTIVE" {
 				if secretList[1].LastUpdated > secretList[0].LastUpdated {
 					clientSecret = secretList[1].ClientSecret
 				} else {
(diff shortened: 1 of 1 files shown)


okta/data_source_okta_app_oauth.go
```diff
@@ -195,6 +202,35 @@ func dataSourceAppOauthRead(ctx context.Context, d *schema.ResourceData, m inter
 	if err != nil {
 		return diag.Errorf("failed to list OAuth's app groups and users: %v", err)
 	}
+	skipClientSecrets := false // Do we ever need to skip doing this?
```
dkulchinsky, line 205: there's an `omit_secret` attribute in the resource, we could follow the same pattern here?

https://github.com/okta/terraform-provider-okta/blob/090c4743242013bed8a5f612c639eacc798f5374/okta/resource_okta_app_oauth.go#L146-L152

https://github.com/okta/terraform-provider-okta/blob/090c4743242013bed8a5f612c639eacc798f5374/okta/resource_okta_app_oauth.go#L532-L534

dkulchinsky:
Hey @monde, do you think you could get this added? though looks like needs a rebase after #1285 was reverted

monde:
@dkulchinsky I'll try to address this in a release this week. Last week my focus was pulled away on to another Okta project.

monde:
@rickardp @dkulchinsky I'll do a separate PR and bring in this behavior.

PR #1285, opened 2022-09-02
oauth app data source: allow to retrieve client_secret

dkulchinsky:
test seem to be ok:
```
=== RUN   TestAccOktaDataSourceAppOauth_read
--- PASS: TestAccOktaDataSourceAppOauth_read (46.13s)
```

but I don't have a huge confidence here, so would appreciate any feedback

/cc @monde


monde (approved):
Looks good to me @dkulchinsky 
`OAuth 2.0 client secret string (used for confidential clients)` and will present on the read for oauth clients https://developer.okta.com/docs/reference/api/apps/#add-oauth-2-0-client-application

PR #1310, opened 2022-09-19
Correct okta_email_customization docs

lucascantor:
Add missing arguments originally listed as attributes

Add example `template_name` argument values for `okta_email_customization` resources

### DIFF
--- a/website/docs/r/email_customization.html.markdown
+++ b/website/docs/r/email_customization.html.markdown
@@ -77,9 +77,5 @@ resource "okta_email_customization" "forgot_password_en_alt" {
 
 ## Attributes Reference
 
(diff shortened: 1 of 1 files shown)


lucascantor:
Perhaps I'm just confused, but for the `is_default` argument, the existing warning to `only set to true for updating a resource` doesn't appear to be valid.

I was able to set `is_default` to true when creating a new `okta_email_customization` resource without encountering any errors.

monde:
Thanks @lucascantor . I cleaned up the attributes list further. Only ID and Links are attributes, thus read only.

monde (approved):

monde:
Released https://github.com/okta/terraform-provider-okta/releases/tag/v3.37.0

PR #1348, opened 2022-11-02
updates to documentation

zlitberg:
- correct that description is required on auth_server_policy
- add clarification and default values on auth_server_policy_rule and app_user_schema_property


monde (approved):
Thank you @zlitberg

Issue #636, opened 2021-09-09
SAML App Provisioning API Integration


drewmullen:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

It would be great if we could pass API credentials directly to the saml app configuration. currently this is a manual step:

![image](https://user-images.githubusercontent.com/3812162/132686727-c3abbeb6-7d98-43d3-aa52-7a685db1f512.png)

### New or Affected Resource(s)

- okta_app_saml

### Potential Terraform Configuration

```hcl
api_integration {
  access_key = ""
  secret_key  = ""
}
```

_secret_key_ should be marked as `sensitive`

### References

Seems like this is a known limitation but i didnt see a FR for it so /shrug: https://registry.terraform.io/providers/okta/okta/latest/docs/resources/app_saml#features

bogdanprodan-okta:
Hi @drewmullen! Thanks for submitting this issue! I'll check whether all the operations are available via public API. If they are I'll add them to the provider. Cheers!

drewmullen:
@bogdanprodan-okta thanks for your update! were you, by chance, able to find an API endpoint? we're looking at making the commands via automation while waiting until the tf provider to be updated - if possible

exitcode0:
looks like this is the endpoint a resource would need to use here
[set-default-provisioning-connection-for-application](https://developer.okta.com/docs/reference/api/apps/#set-default-provisioning-connection-for-application)
The edge case here would be applications that use Oauth for provisioning, this would still require manual click-ops setup in the okta admin console

jackson-theisen:
> looks like this is the endpoint a resource would need to use here [set-default-provisioning-connection-for-application](https://developer.okta.com/docs/reference/api/apps/#set-default-provisioning-connection-for-application) The edge case here would be applications that use Oauth for provisioning, this would still require manual click-ops setup in the okta admin console

[...]

PR #1362, opened 2022-11-10
DRAFT: add support for saml app default provisioning connection

msvechla:
_This PR is still work in progress!_

This PR implements the default provisioning connection for saml / org2org apps: https://developer.okta.com/docs/reference/api/apps/#set-default-provisioning-connection-for-application

See also: https://github.com/okta/terraform-provider-okta/issues/636

The PR adds the following resource block to the `okta_app_saml` resource:

```hcl
  default_provisioning_connection {
      activate = true
      profile {
          auth_scheme = "TOKEN"
          token = "helloworld"
      }
  }
```

What do you think? Is using an in-line resource block for the `okta_app_saml` resource the preferred approach? Or should we move this to a separate resource?

TODO:

- [x] create resource schemas
- [x] implement create
- [ ] implement delete
- [ ] implement update
- [ ] implement import

### DIFF
--- a/okta/resource_okta_app_saml.go
+++ b/okta/resource_okta_app_saml.go
@@ -13,10 +13,12 @@ import (
 )
 
 const (
-	postBinding     = "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST"
-	redirectBinding = "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect"
-	saml11          = "1.1"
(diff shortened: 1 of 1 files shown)


monde:
Okta internal reference: https://oktainc.atlassian.net/browse/OKTA-563715

Thanks @msvechla let me know what kind of collaboration you are looking for here. I can pair with you if you like and I'd prefer this to include a test.

PR #1407, opened 2022-12-19
Add missing arg to docs.

robgero:
The supported argument `user_type` for resource `okta_app_user_base_schema_property` is missing from the docs.

This PR adds the argument to the docs.

This is my first PR here so I'm unsure if there's more you need me to do. I read the contributing docs, and I think this should be enough, please let me know if there is more you need.


monde (approved):

monde:
Thanks @robgero

monde:
Update released in v3.40.0 https://github.com/okta/terraform-provider-okta/releases/tag/v3.40.0

Issue #1493, opened 2023-03-17
oauth okta.networkZones support


cvirtucio:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

The provider's [list of valid scopes](https://github.com/okta/terraform-provider-okta/blob/d0c5948d850d32ae733e756631f198a7602da0a9/okta/resource_okta_app_oauth_api_scope.go#L14-L40) currently does not support the `okta.networkZones.*` scopes, listed [here](https://developer.okta.com/docs/api/oauth2/). This results in an error whenever we try to add it in:

```txt
[...]

PR #1494, opened 2023-03-17
Syncing accepted oauth api scopes with documented options

awagneratzendesk:
This pull request addresses https://github.com/okta/terraform-provider-okta/issues/1493 and syncs the validScopes for oauth api scopes with the options available in the Okta documentation available [here](https://developer.okta.com/docs/api/oauth2/)

This was tested locally to validate that the additional scopes can be successfully applied to an okta_app_oauth_api_scope resource. 

Tests pass successfully and I don't think this requires any additional documentation updates.


monde:
Thanks @awagneratzendesk I'd love to get rid of the provider validator that uses that resource, the API already does input validation. I think we are doing such in v4

monde (approved):
THanks @awagneratzendesk :shipit:

monde (changes requested):
@awagneratzendesk 🤔 "tested locally to validate"

awagneratzendesk:
@monde Sorry about that, was a copy/paste error from moving to my forked repo instead of the main one. Fixed now!

monde (approved):

PR #1495, opened 2023-03-18
New data sources: okta_device  okta_devices

pputman12:
Creates a device and devices data source,

Devices data source has the ability to search with user_id through using "expand" functionality and parsing expanded urls, and with search parameter, which uses https://developer.okta.com/docs/reference/api/devices/#list-devices to search

The intended use case for us, is to get the devices recognized by okta, and push certain fields (such as serial number) into custom attributes in the user profile.  That way we can push those profiles via SCIM to other applications


monde:
Thanks @pputman12 , our PM will be checking this out. In the meantime can you sign our Contributor License Agreement as this is net new code in the provider. https://developer.okta.com/cla/

Thanks

Okta Internal Reference: https://oktainc.atlassian.net/browse/OKTA-592132

pputman12:
Thanks, I signed and emailed it to CLA@okta.com
let me know if you need anything else or any changes.

PR #1498, opened 2023-03-21
add tfplugindocs

duytiennguyen-okta:
We are adding [tfplugindocs](https://github.com/hashicorp/terraform-plugin-docs) to be used as a default docs generator from now on


monde (approved):
:shipit:

PR #1500, opened 2023-03-22
update to v3 and move v2 to local sdk

duytiennguyen-okta:
We are upgrading from v2 to v3 of okta/okta-sdk-golang. Our strategy here is to avoid changing all sdk references at one time, putting us at risk of bugs, is to vendor local v2 and slowly remove those files as we switch over to v3.


monde (approved):
image == fun times

:shipit:

PR #1512, opened 2023-03-29
support brands with sdk v3

duytiennguyen-okta:
switching brands api from v2 to v3


okta/config.go
```diff
@@ -182,6 +190,99 @@ func oktaSDKClient(c *Config) (client *sdk.Client, err error) {
 	return
 }
 
+// TODU remove err???
```
monde: If the method doesn't need to return an error let's remove the error as a return parameter and remove the TODO statement.

okta/config.go
```diff
@@ -182,6 +190,99 @@ func oktaSDKClient(c *Config) (client *sdk.Client, err error) {
...
+		disableHTTPS = strings.HasPrefix(orgUrl, "http://")
+	} else {
+		orgUrl = fmt.Sprintf("https://%v.%v", c.orgName, c.domain)
+	}
+
+	setters := []okta.ConfigSetter{
+		okta.WithOrgUrl(orgUrl),
+		okta.WithCache(false),
+		// TODU
+		// okta.WithHttpClientPtr(httpClient),
```
monde: Why aren't we setting the httpClient in the Okta SDK v3 client set up?

go.mod
```diff
@@ -12,13 +13,15 @@ require (
 	github.com/hashicorp/go-retryablehttp v0.7.2
 	github.com/hashicorp/terraform-plugin-docs v0.14.1
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.25.0
-	github.com/okta/okta-sdk-golang/v2 v2.17.0
+	github.com/kelseyhightower/envconfig v1.4.0
+	github.com/okta/okta-sdk-golang/v3 v3.0.1
```
monde: I would prefer that we do an update on okta sdk golang v3.0.2 to not have it generate `Get` on all the getter names as we discussed. As the v3 hasn't be been marked as the latest release over the v2 series on the GH releases setting I think it is ok to have a large change of method names on a point release.

monde (changes requested):

okta/config.go
duytiennguyen-okta: Because there is not a Config Setter for that. I am updating that in v3.0.2 of Golang sdk, along with the remove the Get in Getter

go.mod
duytiennguyen-okta: See above comment

monde (approved):
:shipit:

PR #1513, opened 2023-03-29
Remove unecessary local input validation using ValidateDiagFunc

monde:
Let Okta API validate input, don't do it artificially with TF SDK `ValidateDiagFunc`'s. Run `gofumpt -w -extra .`

Ref: OKTA-532544

On MAX org

```
 ± grep -h "^... \(PASS\|FAIL\|SKIP\): " /tmp/testacc.log |  awk '{ print $2}' | sort | uniq --count
  16 FAIL:
 213 PASS:
  29 SKIP:
```


duytiennguyen-okta (approved):

ktham:
@monde It's good that we have the API handle the validation to avoid duplication of logic and maintaining that logic in 2 places, but will the validation still happen during terraform "plan" time? I hope it doesn't wait until "apply" time before validations kick in... that would be an undesired regression imho as a user of the okta provider.

monde:
@ktham - @duytiennguyen-okta @jefftayor-okta and I going to take a second look at this

Issue #1338, opened 2022-10-25
v4.0.0 release plan


monde:
I just wanted to give everyone a heads up on our next planned major release, v4.0.0. The release will be focused in this manner:

* CI Release Testing
* Deprecated Resources and Data Sources

## Background

Back in June I opened the sticky issue  [Version v4.0.0 wish list #1183](https://github.com/okta/terraform-provider-okta/issues/1183). That is now actually the v5.0.0 wish list.

I'll be working towards this v4.0.0 release in Okta's Q4 (November, December 2022, January 2023). Hopefully the release will occur early in the quarter. The release will be about stability for all of the new improvements we are planning for v5.0.0.

## CI Release Testing

The v4.0.0 release will be about making [acceptance tests](https://developer.hashicorp.com/terraform/plugin/sdkv2/testing/acceptance-tests) part of our release CI pipeline. Acceptance tests, which run against a live org making API calls over the wire, can take upwards of 40 minutes to run.  Having acceptance tests guard a release with pass/fail will give us greater confidence to not release inadvertent bugs. This is a complicated problem to solve for these reasons:

* Some tests are for a Classic org
* Some tests are for an OIE org
* Some tests are for snowflake situations (combinations of Org feature flags)
* Some tests should never run for release CI purposes
* Release CI should be:
  * Fast
  * Not have any tests that fail intermittently / randomly (no flapping tests allowed)

I've spiked using [VCR](https://github.com/dnaeon/go-vcr) to bring speed and stability to our ACC Tests. This will allow replaying tests for many different kinds of orgs for each resource test. My colleague at Auth0 is using VCR in the Auth0 TF ACC tests as well.

## Deprecated Resources and Data Sources

The v4.0.0 release will also be about removing all deprecated arguments/attributes, and all deprecated resources and data sources.

### Misnamed resources

These resources had incorrect names and the names were marked deprecated in April of 2019 in one of the v2.x.x releases. The resource's behavior was moved to the established resource name. The deprecated names will no longer be recognized.
| deprecated resource name | established resource name |
|-----------------|----------------------|
| `okta_idp`                       | `okta_idp_oidc` |
| `okta_saml_idp`                  | `okta_idp_saml` |
| `okta_saml_idp_signing_key`      | `okta_idp_saml_key` |
| `okta_social_idp`                | `okta_idp_social` |
| `okta_bookmark_app`              | `okta_app_bookmark` |
| `okta_saml_app`                  | `okta_app_saml` |
[...]

PR #1532, opened 2023-04-18
remove deprecated resource and datasource

duytiennguyen-okta:
remove deprecated resource and data source as noted in https://github.com/okta/terraform-provider-okta/issues/1338
I also removed some of attribute that was removed in the resource but not the data source ([data_source_okta_user.go](https://github.com/okta/terraform-provider-okta/pull/1532/files#diff-1370d9b45a7d38f8d662649cf9d364d159c0bc0d698c98ecb862f229b4828da2))
These 2 test still fail `TestAccOktaPolicyRuleIdpDiscovery_crud` and `TestAccOktaUser_customProfileAttributes`


okta/resource_okta_policy_rule_idp_discovery_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 )
 
+// TODU
```
monde, line 10: Is there still a TODO here?

okta/resource_okta_user_test.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 )
 
+// TODU
```
monde, line 18: Done?

monde (approved):
What a liberating PR! `+85 −2,382`
:shipit:

virgofx:
Super excited about removing all the cruft!

virgofx:
Hey @duytiennguyen-okta Any info on when we can expect a 4.x release? Would be happy to test.

duytiennguyen-okta:
@virgofx It should be release by the end of the week

PR #1559, opened 2023-05-17
add engine type check during config phase

duytiennguyen-okta:
Add the engine check to determine if there is a need to get or assign the default access policy (which is OIE only)

### DIFF
--- a/okta/provider.go
+++ b/okta/provider.go
@@ -418,6 +418,11 @@ func providerConfigure(ctx context.Context, d *schema.ResourceData) (interface{}
 		return nil, diag.Errorf("[ERROR] invalid configuration: %v", err)
 	}
 
+	// Discover if the Okta Org is Classic or OIE
+	if org, _, err := config.supplementClient.GetWellKnownOktaOrganization(ctx); err == nil {
+		config.classicOrg = (org.Pipeline == "v1") // v1 == Classic, idx == OIE
+	}
+
 	return &config, nil
 }
 


okta/provider.go
```diff
@@ -418,6 +418,11 @@ func providerConfigure(ctx context.Context, d *schema.ResourceData) (interface{}
 		return nil, diag.Errorf("[ERROR] invalid configuration: %v", err)
 	}
 
+	// Discover if the Okta Org is Classic or OIE
+	if org, _, err := config.supplementClient.GetWellKnownOktaOrganization(ctx); err == nil {
+		config.classicOrg = (org.Pipeline == "v1") // v1 == Classic, idx == OIE
+	}
+
```
monde, line 425: Are you sure that this plays with VCR correctly, the way that you've established VCR in the project? From earlier when I was spiking VCR:
```diff
diff --git a/okta/provider.go b/okta/provider.go
index 1a7af706..01afe62b 100644
--- a/okta/provider.go
+++ b/okta/provider.go
@@ -456,9 +456,16 @@ func providerConfigure(ctx context.Context, d *schema.ResourceData) (interface{}
                return nil, diag.Errorf("[ERROR] invalid configuration: %v", err)
        }

-       // Discover if the Okta Org is Classic or OIE
-       if org, _, err := config.supplementClient.GetWellKnownOktaOrganization(ctx); err == nil {
-               config.classicOrg = (org.Pipeline == "v1") // v1 == Classic, idx == OIE
+       // NOTE: Don't make this call when VCR is playing/recording as it will occur
+       // outsite of the VCR transport
+       if os.Getenv("OKTA_VCR_TF_ACC") == "" {
+               // NOTE: validate credentials during initial config with a call to
+               // /api/v1/users/me
+               if config.apiToken != "" {
+                       if _, _, err := config.oktaClient.User.GetUser(ctx, "me"); err != nil {
+                               return err, diag.Errorf("[ERROR] : api_token is not correct %v", err)
+                       }
+               }
        }

        return &config, nil
```

monde (commented):
Would be nice to have some background in the PR description to give context to the meaning of the PR.

okta/provider.go
duytiennguyen-okta: yeah that was moved into [loadAndValidate](https://github.com/okta/terraform-provider-okta/blob/master/okta/config.go#L70)

monde (approved):
:shipit:

PR #1562, opened 2023-05-18
add skip user and skip group

duytiennguyen-okta:
Bringing back `skip_users` and `skip_groups` on apps data sources. They are no-ops and irrelevant in v4.x.x on the apps data sources. But we forgot to mark them deprecated along with the `users` and `groups` properties that were marked deprecated in v3.x.x and removed in v4.x.x .

### DIFF
--- a/okta/app.go
+++ b/okta/app.go
@@ -89,6 +89,23 @@ var (
 		},
 	}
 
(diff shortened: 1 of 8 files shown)


monde (approved):
:shipit:

PR #1586, opened 2023-06-08
allow refresh token removal

duytiennguyen-okta:
Allow refresh_token removal by switching check from refresh_token_leeway to actual check in grant types

### DIFF
--- a/okta/resource_okta_app_oauth.go
+++ b/okta/resource_okta_app_oauth.go
@@ -770,21 +770,31 @@ func buildAppOAuth(d *schema.ResourceData) *sdk.OpenIdConnectApplication {
 	}
 
 	refresh := &sdk.OpenIdConnectApplicationSettingsRefreshToken{}
-	hasRefresh := false
+	var hasRefresh bool
+	for _, grant_type := range grantTypes {
+		if grant_type == refreshToken {
+			hasRefresh = true
+		}
+	}
 	if rotate, ok := d.GetOk("refresh_token_rotation"); ok {
 		refresh.RotationType = rotate.(string)
-		hasRefresh = true
 	}
 
 	if leeway, ok := d.GetOk("refresh_token_leeway"); ok {
 		refresh.LeewayPtr = int64Ptr(leeway.(int))
-		hasRefresh = true
 	}
 
 	if hasRefresh {
 		app.Settings.OauthClient.RefreshToken = refresh
 	}
 
+	// TODU
+	// if !hasRefresh && refresh != nil {
+	// 	return nil, errors.New("does not have refresh grant type but refresh_token_rotation and refresh_token_leeway exist in payload")
+	// }
+
+	// TODU
+	// should we unset here refresh_token_rotation, refresh_token_leeway?
 	app.Visibility = buildAppVisibility(d)
 	app.Accessibility = buildAppAccessibility(d)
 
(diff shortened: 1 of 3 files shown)


okta/resource_okta_app_oauth.go
```diff
@@ -770,21 +770,31 @@ func buildAppOAuth(d *schema.ResourceData) *sdk.OpenIdConnectApplication {
...
 	if leeway, ok := d.GetOk("refresh_token_leeway"); ok {
 		refresh.LeewayPtr = int64Ptr(leeway.(int))
-		hasRefresh = true
 	}
 
 	if hasRefresh {
 		app.Settings.OauthClient.RefreshToken = refresh
 	}
 
+	// TODU
```
duytiennguyen-okta: Should we add an error or a warning here?

okta/resource_okta_app_oauth.go
```diff
@@ -770,21 +770,31 @@ func buildAppOAuth(d *schema.ResourceData) *sdk.OpenIdConnectApplication {
...
+	// if !hasRefresh && refresh != nil {
+	// 	return nil, errors.New("does not have refresh grant type but refresh_token_rotation and refresh_token_leeway exist in payload")
+	// }
+
+	// TODU
+	// should we unset here refresh_token_rotation, refresh_token_leeway?
```
duytiennguyen-okta: Should we unset refresh_token_rotation and refresh_token_leeway from the resource if it does not include the refresh?

okta/resource_okta_app_oauth_test.go
```diff
@@ -87,10 +87,11 @@ func TestAccResourceOktaAppOauth_refreshToken(t *testing.T) {
 	//       Skipping for now assuming that the okta account doesn't have this feature enabled.
 	//       If this feature is enabled or Okta releases this to all this test should be enabled.
 	//       SEE https://help.okta.com/en/prod/Content/Topics/Apps/apps-fbm-enable.htm
-	t.Skip("This is an 'Early Access Feature' and needs to be enabled by Okta, skipping this test as it fails when this feature is not available")
+	// t.Skip("This is an 'Early Access Feature' and needs to be enabled by Okta, skipping this test as it fails when this feature is not available")
```
monde: I'd delete this line

okta/resource_okta_app_oauth.go
```diff
@@ -770,21 +770,31 @@ func buildAppOAuth(d *schema.ResourceData) *sdk.OpenIdConnectApplication {
 	}
 
 	refresh := &sdk.OpenIdConnectApplicationSettingsRefreshToken{}
-	hasRefresh := false
+	var hasRefresh bool
+	for _, grant_type := range grantTypes {
+		if grant_type == refreshToken {
+			hasRefresh = true
```
monde, line 776: there could be a break after this `hasRefresh = true`

okta/resource_okta_app_oauth.go
monde: if it should have have a refresh value, but does not, that is an error is it not?

okta/resource_okta_app_oauth_test.go
duytiennguyen-okta: So delete the skip or not?

monde (approved):

PR #1600, opened 2023-06-19
Update the description for the "platform_include" block of "app_signon_policy_rule" to outline requirement for the "os_expression" argument to be set when "os_type" is set to "OTHER" 

achuchulev:
Currently, the Okta provider requires the "os_expression" argument to be set when "os_type" is set to "OTHER" which is not clearly outlined within the documentation and when omitted, Terraform will return the following error:

`Error: failed to create app sign on policy rule: the API returned an error: Api validation failed: conditions.platform. Causes: errorSummary: conditions.platform: The condition 'platform' must have os.expression if os.type is OTHER.`

PS.
This could be reflected in the complex example configuration as well.


monde (approved):
Thank you @achuchulev !

PR #1606, opened 2023-06-25
New data source: okta_group_rule

steveAG:
Adds a new data source to retrieve a group rule.

We're using this to build an optional safety check in our application module to reduce the impact of unexpected mass unassign from a high-impact application.

- Read the current rule state (new data source) and compare it to the provided expression input variable to detect an expression change.
- Use a resource precondition to block apply on expression change if user de-provisioning is enabled for the associated application (read from the existing application data source).

### DIFF
--- a/examples/okta_group_rule/datasource.tf
+++ b/examples/okta_group_rule/datasource.tf
@@ -1,0 +1,15 @@
+resource "okta_group" "test" {
+  name = "testAcc_replace_with_uuid"
+}
+
+resource "okta_group_rule" "test" {
+  name              = "testAcc_replace_with_uuid"
+  status            = "ACTIVE"
+  group_assignments = [okta_group.test.id]
+  expression_type   = "urn:okta:expression:1.0"
+  expression_value  = "String.startsWith(user.firstName,\"andy\")"
+}
+
+data "okta_group_rule" "test" {
+  id          = okta_group_rule.test.id
+}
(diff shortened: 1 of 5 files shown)


exitcode0:
As an unrelated note, I'd love to hear more about this module you've built

monde:
I like this PR thanks @steveAG . I'm playing tag with @duytiennguyen-okta to write an ACC test for it.

monde:
@steveAG we are taking this PR and @duytiennguyen-okta is going to write an ACC test for it. But we need to you to send in an Okta Individual Contributor License Agreement
* https://developer.okta.com/cla/
* see link to PDF "sign a CLA" at the bottom of that page
Thanks again

steveAG:
Thanks @monde, I signed & emailed the CLA on Sunday when submitting the PR because I saw it called out on other PRs, so we should be all set on that front.

@exitcode0 Sorry I'm not ignoring the above request, I think I found a cleaner approach that solves some edge cases and doesn't depend on our org-specific metadata, so I'm doing a bit of a re-write to test that and should have some details tomorrow/Sat depending on whether that approach works. To be clear this doesn't impact the PR.

monde:
> Thanks @monde, I signed & emailed the CLA on Sunday when submitting the PR because I saw it called out on other PRs, so we should be all set on that front.

I see it, apologies for pinging you about unnecessarily.

monde:
Brought this via https://github.com/okta/terraform-provider-okta/pull/1617 it will be in the next release. Thanks @steveAG , thanks @exitcode0

steveAG:
Awesome, thanks @monde!

PR #1610, opened 2023-06-26
prep release v4.0.3

duytiennguyen-okta:


monde (approved):

tmatilai:
Thanks for the release, @duytiennguyen-okta !

But would it be possible to follow the SemVer standard for version numbers? For example to bump the minor when there are new features.
That way the users can use `~> 4.1` kind of requirements when depending on a feature, but allowing all future versions until next major (which could break the module).

monde:
Good suggestion @tmatilai . Getting SemVer dialed in on Okta TF provider has been a moving target at times.

PR #1327, opened 2022-10-07
Fix/refine okta_app_oauth behavior for args `auto_key_rotation`, `pkce_required`, `token_endpoint_auth_method`

monde:
Fix/refine okta_app_oauth behavior w.r.t. API documentation for
    arguments `auto_key_rotation`, `pkce_required`,
    `token_endpoint_auth_method`

Closes #1325

PR #1626, opened 2023-07-07
removed suppression of pkce_required

cvirtucio:
# Description

[PR #1327](https://github.com/okta/terraform-provider-okta/pull/1327) added logic to set a default value of `true` on `pkce_required` when the app type is `browser` or `client`. However, it also added a suppression function that suppresses the change if that default value was set, and the caller decides to explicitly set it to `false` later on. This change removes the function, so that callers can explicitly set it to `false` and see the diff.

I considered gating the original suppression logic on `browser` and `client` apps, but I'm not entirely sure there's much benefit, as it still wouldn't solve the problem of not seeing the diff when `pkce_required` is explicitly set to `false` after the default value of `true` is applied for `browser` and `client` apps.

# Expectations

## Expected Outcome

When terraform similar to this is used:

```hcl
resource "okta_app_oauth" "foobarbaz" {
  label                      = "foo"
  type                       = "browser"
}
```

and applied, setting `pkce_required` to `true`, and we later change it to this:

```hcl
resource "okta_app_oauth" "foobarbaz" {
  label                      = "foo"
  type                       = "browser"
  pkce_required = false
}
```

the change should show up in when we run `terraform plan`.

## Actual Outcome

The change does not appear when we run `terraform plan`.


tgoodsell-tempus (changes requested):
FYI, I think the original PR had added this to try and make the various "upstream defaults" the Okta API has work correctly. I think the "better" answer would be the use of a `DefaultFunc` to replace this logic, set based on the value of `type`.

The okta logic from their api docs linked is:
```
Use pkce_required to require PKCE for your confidential clients using the [Authorization Code flow](https://developer.okta.com/docs/guides/implement-grant-type/authcodepkce/main/). If token_endpoint_auth_method is none, pkce_required needs to be true. If pkce_required isn't specified when adding a new application, Okta sets it to true by default for browser and native application types.
```

If it doesn't exist already I think a validate/validate function can be used to require this to be true when `endpoint_auth_method` is `none`, if you wanted to try that as well.

cvirtucio:
> FYI, I think the original PR had added this to try and make the various "upstream defaults" the Okta API has work correctly. I think the "better" answer would be the use of a `DefaultFunc` to replace this logic, set based on the value of `type`.
> 
> The okta logic from their api docs linked is:
> 
> ```
> Use pkce_required to require PKCE for your confidential clients using the [Authorization Code flow](https://developer.okta.com/docs/guides/implement-grant-type/authcodepkce/main/). If token_endpoint_auth_method is none, pkce_required needs to be true. If pkce_required isn't specified when adding a new application, Okta sets it to true by default for browser and native application types.
> ```
> 
> If it doesn't exist already I think a validate/validate function can be used to require this to be true when `endpoint_auth_method` is `none`, if you wanted to try that as well.

I looked into that, and it doesn't seem like there's a way to access other parameters from within the `DefaultFunc` of one parameter.

tgoodsell-tempus:
> > FYI, I think the original PR had added this to try and make the various "upstream defaults" the Okta API has work correctly. I think the "better" answer would be the use of a `DefaultFunc` to replace this logic, set based on the value of `type`.
> > The okta logic from their api docs linked is:
> > ```
> > Use pkce_required to require PKCE for your confidential clients using the [Authorization Code flow](https://developer.okta.com/docs/guides/implement-grant-type/authcodepkce/main/). If token_endpoint_auth_method is none, pkce_required needs to be true. If pkce_required isn't specified when adding a new application, Okta sets it to true by default for browser and native application types.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If it doesn't exist already I think a validate/validate function can be used to require this to be true when `endpoint_auth_method` is `none`, if you wanted to try that as well.
> 
> I looked into that, and it doesn't seem like there's a way to access other parameters from within the `DefaultFunc` of one parameter.

@cvirtucio Yea good shout, I did another round of research and I'm pretty sure it would just have to go into the `CustomizeDiff` block instead, see: https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/customizing-differences, but that isn't perfect.

However, I do think I have the "ideal" solution. In this case you would just want to set the `Computed` flag to true, and ensure that we're writing back the value to the schema on the Create, Read, and Update operations. Since the API docs call out that Okta sets the value for us when not provided, then we can use the computed flag to indicate that behavior and not need the diff suppress. 

If you also wanted to be "nice" and go a step further, you could also try and shim a validation in using the `CustomizeDiff` func, for ensuring it's set when `endpoint_auth_method` is none. However, since the docs indicate this is all hacky, not properly implemented (https://github.com/hashicorp/terraform-plugin-sdk/issues/233), it may be wiser to not include this and wait till this gets moved to the new SDK, if the written logic doesn't work cleanly and consistently.

cvirtucio:
> > > FYI, I think the original PR had added this to try and make the various "upstream defaults" the Okta API has work correctly. I think the "better" answer would be the use of a `DefaultFunc` to replace this logic, set based on the value of `type`.
> > > The okta logic from their api docs linked is:
> > > ```
> > > Use pkce_required to require PKCE for your confidential clients using the [Authorization Code flow](https://developer.okta.com/docs/guides/implement-grant-type/authcodepkce/main/). If token_endpoint_auth_method is none, pkce_required needs to be true. If pkce_required isn't specified when adding a new application, Okta sets it to true by default for browser and native application types.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > If it doesn't exist already I think a validate/validate function can be used to require this to be true when `endpoint_auth_method` is `none`, if you wanted to try that as well.
> > 
> > 
> > I looked into that, and it doesn't seem like there's a way to access other parameters from within the `DefaultFunc` of one parameter.
> 
> @cvirtucio Yea good shout, I did another round of research and I'm pretty sure it would just have to go into the `CustomizeDiff` block instead, see: https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/customizing-differences, but that isn't perfect.
> 
> However, I do think I have the "ideal" solution. In this case you would just want to set the `Computed` flag to true, and ensure that we're writing back the value to the schema on the Create, Read, and Update operations. Since the API docs call out that Okta sets the value for us when not provided, then we can use the computed flag to indicate that behavior and not need the diff suppress.
> 
> If you also wanted to be "nice" and go a step further, you could also try and shim a validation in using the `CustomizeDiff` func, for ensuring it's set when `endpoint_auth_method` is none. However, since the docs indicate this is all hacky, not properly implemented ([hashicorp/terraform-plugin-sdk#233](https://github.com/hashicorp/terraform-plugin-sdk/issues/233)), it may be wiser to not include this and wait till this gets moved to the new SDK, if the written logic doesn't work cleanly and consistently.

made `pkce_required` a `computed` attribute and added some additional checks based on the API docs: ﻿c201537faa2593eef08e67f2a517f3528ba70586

monde:
@cvirtucio @tgoodsell-tempus you all good with this PR now? I'm going to bring it in on another PR and write some additional tests.

monde:
Okta internal reference https://oktainc.atlassian.net/browse/OKTA-671285

cvirtucio:
> @cvirtucio @tgoodsell-tempus you all good with this PR now? I'm going to bring it in on another PR and write some additional tests.

@monde  nothing further from my end.

PR #1640, opened 2023-07-20
remove errant space in URL target in CHANGELOG

thekbb:
The changelog fails to render properly due to an errant space. 
screenshot of error in CHANGELOG


thekbb:
This is fixed in de2233b

PR #1677, opened 2023-08-24
add support for brand creation

duytiennguyen-okta:
add support for brand creation and convert brand resource to plugin framework
Update resource to match new sdk attribute


website/docs/r/brand.html.markdown
```diff
@@ -8,54 +8,57 @@ description: |-
...
-  brand_id = "default"
-  agree_to_custom_privacy_policy = true
-  custom_privacy_policy_url      = "https://example.com/privacy-policy"
-  remove_powered_by_okta         = true
+  name = "example
 }
 ```
 
+## Attention
+Due to the way brand api works, you will need to first create a brand with only the name, then you can update other attirbutes later. Failure to do so will cause terraform to error out
```
monde: Is that statement valid? It appears a true create api call is made in the resource's create method: https://github.com/okta/terraform-provider-okta/pull/1677/files#diff-eeba2406217f346ad9992036218a37f2d1aab97a73b1f4d1580a6ac5eed371e5R166

okta/resource_okta_email_domain.go
```diff
@@ -104,7 +99,6 @@ func resourceEmailDomainRead(ctx context.Context, d *schema.ResourceData, m inte
 		arr[i] = map[string]interface{}{
 			"fqdn":        dnsValidation[i].GetFqdn(),
 			"record_type": dnsValidation[i].GetRecordType(),
-			"expiration":  dnsValidation[i].GetExpiration(),
```
monde, line 107: Why is `expiration` removed? The need for the removal is not documented in a commit message or in the PR description.

monde: Also, the PR title is "add support for brand creation" yet this is modifying okta email domain resource.

monde: Maybe `GetExpiration()` was removed form okta-sdk-golang v3? Expiration is still documented in the API so it seems like a flaw to remove it from the sdk and the TF provider https://developer.okta.com/docs/api/openapi/okta-management/management/tag/CustomDomain/#tag/CustomDomain/operation/createCustomDomain

okta/resource_okta_device_assurance_policy_android_os.go
```diff
@@ -296,9 +296,9 @@ func buildDeviceAssuranceAndroidPolicyRequest(model policyDeviceAssuranceAndroid
 	android.SetName(model.Name.ValueString())
 	android.SetPlatform(okta.PLATFORM_ANDROID)
 	if len(model.DiskEncryptionType) > 0 {
-		diskEncryptionType := make([]okta.DiskEncryptionType, 0)
+		diskEncryptionType := make([]okta.DiskEncryptionTypeAndroid, 0)
```
monde, line 299: I can see that the need to change to the *_os.go files is from okta-sdk-golang interfaces changing. However, this should should still be noted in the PR description.

monde (changes requested):
I like the the brand refactor to TF plugin framework, nice work. I'm concerned about the expiration argument that pointed on the email domain resource. Also, additions to the PR description will be good to have both for code review now, and for bug hunting in the future.

website/docs/r/brand.html.markdown
monde:
```go
createdBrand, _, err := r.oktaSDKClientV3.CustomizationApi.CreateBrand(ctx).CreateBrandRequest(reqBody).Execute()
```

monde (approved):

okta/resource_okta_device_assurance_policy_android_os.go
duytiennguyen-okta: Change in go sdk

okta/resource_okta_email_domain.go
duytiennguyen-okta: Change in go sdk

PR #1685, opened 2023-08-30
add brand id to okta_domain

duytiennguyen-okta:
add brand id to okta_domain

### DIFF
--- a/examples/okta_domain/basic.tf
+++ b/examples/okta_domain/basic.tf
@@ -1,3 +1,4 @@
 resource "okta_domain" "test" {
-  name   = "testAcc-replace_with_uuid.example.com"
+  name                    = "testAcc.example.edu"
+  certificate_source_type = "MANUAL"
 }
--- a/okta/resource_okta_domain.go
+++ b/okta/resource_okta_domain.go
@@ -6,7 +6,7 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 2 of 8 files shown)


examples/okta_domain/basic.tf
```diff
@@ -1,3 +1,4 @@
 resource "okta_domain" "test" {
-  name   = "testAcc-replace_with_uuid.example.com"
```
duytiennguyen-okta, line 2: The endpoint no longer allow domain to be named testAcc-randomnumber.example.com

okta/resource_okta_domain.go
```diff
@@ -71,49 +76,63 @@ func resourceDomain() *schema.Resource {
...
 		return diag.Errorf("failed to create domain: %v", err)
 	}
-	d.SetId(domain.Id)
+	d.SetId(domain.GetId())
+	if brandId, ok := d.GetOk("brand_id"); ok {
+		_, _, err = getOktaV3ClientFromMetadata(m).CustomDomainApi.ReplaceCustomDomain(ctx, d.Id()).UpdateDomain(okta.UpdateDomain{BrandId: brandId.(string)}).Execute()
+		if err != nil {
+			return diag.Errorf("failed to update domain: %v", err)
+		}
+		_, err = getOktaV3ClientFromMetadata(m).CustomizationApi.DeleteBrand(ctx, domain.GetBrandId()).Execute()
```
duytiennguyen-okta, line 93: Because whenever a custom domain is created, a brand will be created along with that custom domain. If the customer is using another brand right away, it should be deleted

monde (approved):

PR #1691, opened 2023-08-31
Internal API calls guards during OAuth 2.0 authentication

monde:
Guards against making internal API calls when the SDK client is using OAuth 2.0 authentication have been put in place.

Resources
- `okta_profile_mapping`
- `okta_app_oauth`

Note that resource `okta_security_notification_emails` and `okta_rate_limiting` are OAuth 2.0 incompatible.

### DIFF
--- a/okta/config.go
+++ b/okta/config.go
@@ -219,6 +219,10 @@ func (c *Config) IsClassicOrg(ctx context.Context) bool {
 	return c.classicOrg
 }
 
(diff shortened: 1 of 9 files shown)


duytiennguyen-okta (approved):

monde:
Okta internal reference: https://oktainc.atlassian.net/browse/OKTA-644888

PR #1697, opened 2023-09-06
Update stale.yml

exitcode0:
`waiting-response` should be removed if the author of an issue or PR responds on the PR

Tweak the `time-to-stale` and `time-to-close`
Total `time-to-close` remains the same (65 days)
 - `Stale` label will be applied sooner (60 Days -> 30 Days)
 - Users will have more time between `Stale` label being added and issue being closed (5 days -> 35 days)
   -  this better accounts for people taking extended PTO from their day jobs which may mean they miss the notification from github of the `Stale` label being added to their issue


monde (approved):
Make it so.
Thanks for putting some thought and giving good suggestions here @exitcode0

PR #1705, opened 2023-09-09
add tf plugin docs template

duytiennguyen-okta:
Add tf plugin docs template to generate docs with example


MikeMondragon-okta (approved):
wow

monde:

PR #1715, opened 2023-09-12
Support EC Keys Quick Fix

tgoodsell-tempus:
This should enable support for EC keys per Okta's API documentation in the quick/dirty format. This isn't my ideal long term solution, nor does it address 
potential issues such as the lack of the `alg`, `use`, and `crv` fields in the JWKS documents, nor support for future JWKS formats.

- add the x and y attributes for jwks blocks
- add EC test client and format terraform resources
- update test EC key to something valid
- update resource name and add to oauth test config


monde:
I had to bring this in on a branch/PR that I owned. Did that and added some clean ups.

PR #1716, opened 2023-09-12
Chore: Format Terraform Files

tgoodsell-tempus:
Formatting the Terraform files here, so they're in a decent shape and future format runs don't try to fix all of these along with the target files.

### DIFF
--- a/examples/okta_admin_role_custom_assignments/basic.tf
+++ b/examples/okta_admin_role_custom_assignments/basic.tf
@@ -1,5 +1,5 @@
 variable "hostname" {
-   type=string
+  type = string
 }
 
 locals {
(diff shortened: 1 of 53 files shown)


monde:
@tgoodsell-tempus did you do this by hand (😲) or with some tf config linter/formatter (😎)

monde (approved):
Awesome, thank you @tgoodsell-tempus

monde:
I need to cherry pick this into a branch I own due to branch merging protections, fyi

tgoodsell-tempus:
> @tgoodsell-tempus did you do this by hand (😲) or with some tf config linter/formatter (😎)

@monde I use a alias (I'm a `zsh` user) such as:
```zsh
def tfmt() {
    terraform fmt --recursive
}
```

I've a few other combo aliases I use when commiting code and so on to make sure that's run when I'm working on stuff like this. Would highly recommend from a workflow standpoint:

For example, I have a commit alias which does:
```zsh
terraform fmt --recursive
git add -A
git commit -m "$1" (message from my shell)
git push -u origin
```

and so on

PR #1725, opened 2023-09-14
Client OAuth with PKCS#1 or PKCS#8 private key

monde:
Makes the provider be flexible with private key format. okta-sdk-golang v3 needs to be updated as well. Use v2 source in this PR as a guide.


duytiennguyen-okta (approved):

monde:

Issue #1582, opened 2023-06-06
DataSource `/api/v1/org`


exitcode0:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description
Would be good to have a datasource for this api endpoint as this is now being used in ORNs which are in turn used in `okta_resource_set`

### New or Affected Resource(s)

- data okta_org
- data okta_instance

### Potential Terraform Configuration

```hcl
# Workaround config
locals {
  OKTA_ORG_ID  = jsondecode(data.http.OKTA_ORG_ID.body).id
}
data "http" "OKTA_ORG_ID" {
  url             = "${local.instance_url}/api/v1/org"
  request_headers = { Authorization = "SSWS ${var.OKTA_API_TOKEN}" }
}
```

```hcl
# proposed config
data "okta_org" "myOrg" {}

output "okta_org_id" {
  value = data.okta_org.myOrg.id
}

locals {
  okta_org_address1              = okta_org.myOrg.address1
  okta_org_address2              = okta_org.myOrg.address2
  okta_org_city                  = okta_org.myOrg.city
  okta_org_country               = okta_org.myOrg.country
  okta_org_endUserSupportHelpURL = okta_org.myOrg.endUserSupportHelpURL
  okta_org_id                    = okta_org.myOrg.id
  okta_org_name                  = okta_org.myOrg.name
  okta_org_phoneNumber           = okta_org.myOrg.phoneNumber
  okta_org_postalCode            = okta_org.myOrg.postalCode
  okta_org_state                 = okta_org.myOrg.state
  okta_org_status                = okta_org.myOrg.status
  okta_org_subdomain             = okta_org.myOrg.subdomain
  okta_org_supportPhoneNumber    = okta_org.myOrg.supportPhoneNumber
}
```

### References

- #0000

exitcode0:
This could also probably be done with `/.well-known/okta-organization`

duytiennguyen-okta:
OKTA internal reference https://oktainc.atlassian.net/browse/OKTA-616650

PR #1726, opened 2023-09-15
Add Org Metadata Data Source

tgoodsell-tempus:
This is intended to emit a data source corresponding to the "well known org metadata" 
(https://developer.okta.com/docs/api/openapi/okta-management/management/tag/OrgSetting/#tag/OrgSetting/operation/getWellknownOrgMetadata). Since it's already used internally in the 
provider, I'm just adding the rest of the necessary structs to make all the data easily accesible, then pumping it into a data source. 

This should be useful for "in terraform" logic we want to use; such as org id, urls, etc; but frankly don't want to involve having to setup root module variables feeding data from env 
vars.

Trying to do this with the `terraform-provider-framework` since thats their desired plugin maker now.

### DIFF
--- a/examples/okta_org_metadata/datasource.tf
+++ b/examples/okta_org_metadata/datasource.tf
@@ -1,0 +1,1 @@
+data "okta_org_metadata" "test" {}
--- a/go.mod
+++ b/go.mod
@@ -55,6 +55,7 @@ require (
 	github.com/hashicorp/logutils v1.0.0 // indirect
 	github.com/hashicorp/terraform-exec v0.19.0 // indirect
 	github.com/hashicorp/terraform-json v0.17.1 // indirect
+	github.com/hashicorp/terraform-plugin-testing v1.5.1 // indirect
 	github.com/hashicorp/terraform-registry-address v0.2.2 // indirect
 	github.com/hashicorp/terraform-svchost v0.1.1 // indirect
 	github.com/hashicorp/yamux v0.0.0-20211028200310-0bc27b27de87 // indirect
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,6 @@ require (
 	github.com/hashicorp/terraform-plugin-log v0.9.0
 	github.com/hashicorp/terraform-plugin-mux v0.12.0
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.29.0
-	github.com/hashicorp/terraform-plugin-testing v1.5.1
 	github.com/kelseyhightower/envconfig v1.4.0
 	github.com/okta/okta-sdk-golang/v3 v3.0.14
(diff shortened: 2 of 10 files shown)


tgoodsell-tempus:
FYI, for the future (as mentioned in my TODOs), the desired state (Hashi Plugin Desired State) is to use https://developer.hashicorp.com/terraform/plugin/framework/handling-data/attributes/single-nested for some of the structured attributes here. 

That requires proto v6 to be used as the server protocol, which means a drop of support for v5 proto. 

I'm using their "workaround" types for this now, however, this (and other things) can be refactored with all the latest and greatest, once Okta is OK with pulling the plug on Terraform CLI support older than 1.0.

exitcode0:
I recall hearing or reading that `https://{yourOktaDomain}/.well-known/okta-organization` days were numbered and the intent was to replace it with `https://{yourOktaDomain}/api/v1/org`

I don't work at Okta though, so I could just be completely wrong ¯\\_(ツ)_/¯ 

Related issue: https://github.com/okta/terraform-provider-okta/issues/1582

tgoodsell-tempus:
Ya fair, we'll need someone ITK to comment on this. Based on their API docs, those two endpoints won't return the same data, so we probably would need both unless that's changing:
https://developer.okta.com/docs/api/openapi/okta-management/management/tag/OrgSetting/#tag/OrgSetting/operation/getWellknownOrgMetadata

https://developer.okta.com/docs/api/openapi/okta-management/management/tag/OrgSetting/#tag/OrgSetting/operation/getOrgSettings

tgoodsell-tempus:
OK this should be pretty "complete" now as far as I'm concerned.

tgoodsell-tempus:
@monde @duytiennguyen-okta Just tagging you to ensure you know I am looking for a review (and eventual approval) on adding this resource.

monde:
@tgoodsell-tempus - @duytiennguyen-okta and I will do a review and give feedback. Thanks for the work.

duytiennguyen-okta:
@tgoodsell-tempus lgtm but why are you using protocolv6 instead of v5?

tgoodsell-tempus:
@duytiennguyen-okta 

For why switching to V6 see: https://developer.hashicorp.com/terraform/plugin/terraform-plugin-protocol#protocol-version-6 (basically some of the better functions in the framework provider are locked behind this).

However this PR I'm not actually switching to V6 directly, all I've done here is instead of initing the framework provider as a V5 plugin, I'm creating it as a V6 plugin then muxing it to a V5 plugin before merging the two. This should not change anything functionally, but I'm doing it + adding the TODO line as a reminder this should be converted to V6 (AKA making this one the native provider and then using V5 to V6 for the older provider) as a documentation/code reminder.

PR #1760, opened 2023-10-11
fix panic

duytiennguyen-okta:
Fix panic when convertInterfaceArrToStringArr

### DIFF
--- a/okta/utils.go
+++ b/okta/utils.go
@@ -147,8 +147,7 @@ func convertInterfaceToStringArr(purportedList interface{}) []string {
 func convertInterfaceArrToStringArr(rawArr []interface{}) []string {
 	arr := make([]string, len(rawArr))
 	for i, thing := range rawArr {
-		a, ok := thing.(string)
-		if ok {
+		if a, ok := thing.(string); ok {
 			arr[i] = a
 		}
 	}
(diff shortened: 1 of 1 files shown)


okta/utils.go
```diff
@@ -238,7 +241,10 @@ func ensureNotDefault(d *schema.ResourceData, t string) error {
 
 func getMapString(m map[string]interface{}, key string) string {
 	if v, ok := m[key]; ok {
-		return v.(string)
+		if res, ok := v.(string); ok {
+			return res
+		}
+		return ""
```
monde: extra there's an extra `return ""`, it would occur on 249 when coming out of the if statement

okta/utils.go
```diff
@@ -147,7 +147,10 @@ func convertInterfaceToStringArr(purportedList interface{}) []string {
 func convertInterfaceArrToStringArr(rawArr []interface{}) []string {
 	arr := make([]string, len(rawArr))
 	for i, thing := range rawArr {
-		arr[i] = thing.(string)
+		a, ok := thing.(string)
```
monde: like line 244 `if a, ok := thing.(string); ok {`

monde (approved):
does the linter also pick up these suggestions?

PR #1794, opened 2023-10-31
Tighten up the data source `okta_group` implementation and documentation

### LINKED ISSUE #1783: datasource_okta_group - `limit` value
### Community Note

### Expected Behavior

We should throw an explicit error if okta_group would return an non-deterministic result

### Actual Behavior

we [set `limit` to `1`](https://github.com/okta/terraform-provider-okta/blob/master/okta/data_source_okta_group.go#L88) rather than 2 and then explicitly check the length after

I expect that this could lead to buggy behavior where two or more groups match the dataSource criteria

### References

- #0000

monde:
Given data source `okta_group`

- Document how Okta API treats a name query as name starts with
- Improve logic of searching for a group 
- Improve error messages when searching for a group
- Includes ACC and VCR cassette for the test

Closes #1783

### DIFF
--- a/okta/data_source_okta_group.go
+++ b/okta/data_source_okta_group.go
@@ -85,21 +85,34 @@ func findGroup(ctx context.Context, name string, d *schema.ResourceData, m inter
 		}
 		group = respGroup
 	} else {
-		searchParams := &query.Params{Q: name, Limit: 1}
+		// NOTE: Okta API query on name is effectively a starts with query, not
+		// an exact match query.
+		searchParams := &query.Params{Q: name}
+
+		// NOTE: Uncertain of the OKTA /api/v1/groups API drifted during Classic
+		// (when this data source was originally created) to OIE migration.
+		// Currently, Okta API enforces unique names on all groups regardless of
+		// type so type is essentially a meaningless parameter in OIE.
 		t, okType := d.GetOk("type")
 		if okType {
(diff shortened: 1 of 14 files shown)


duytiennguyen-okta (approved):

okta/data_source_okta_group.go
```diff
@@ -85,21 +85,34 @@ func findGroup(ctx context.Context, name string, d *schema.ResourceData, m inter
...
 		group = respGroup
 	} else {
-		searchParams := &query.Params{Q: name, Limit: 1}
+		// NOTE: Okta API query on name is effectively a starts with query, not
+		// an exact match query.
+		searchParams := &query.Params{Q: name}
+
+		// NOTE: Uncertain of the OKTA /api/v1/groups API drifted during Classic
+		// (when this data source was originally created) to OIE migration.
+		// Currently, Okta API enforces unique names on all groups regardless of
```
exitcode0, line 94: > Currently, Okta API enforces unique names on all groups regardless of type so type is essentially a meaningless parameter in OIE.

@monde while you can't use the API to create a new group if the name is taken
duplicate group names can still exist via imported `APP_GROUP`s which don't enforce uniqueness
an imported `APP_GROUP` can have the same name as an `APP_GROUP` from the same or different source app, and it can have the same name as an existing `OKTA_GROUP`

PR #1815, opened 2023-11-15
Okta App OAuth: Make Omit Secret Safer To Set

tgoodsell-tempus:
The goal for this PR was:

- Ensure that when we actually do the UPDATE where `omit_secret` is activated, that update does not send the `client_secret` value stored in state to the Okta API.
- Add comments and fix the state setting, to better reflect the reality of what this function is doing
- We also added "no-op" but semantic changes to how the return data from the app calls is used. Basically we use the output value instead of the passed in request, even though this is the same since the lower level API calls use that object to fill in the data from the response. It makes more "sense" from the Golang SDK standpoint to specifically show we're using the results of the API calls to set the state values, instead of the more ambiguous method currently in use.
- Update document to explain more clearly the functions of the `omit_secret` and `client_basic_secret` arguments, as well as the `client_secret` attribute.

### DIFF
--- a/examples/data-sources/okta_default_signin_page/datasource.tf
+++ b/examples/data-sources/okta_default_signin_page/datasource.tf
@@ -1,0 +1,7 @@
+data "okta_brands" "test" {
+}
+
+
+data "okta_default_signin_page" "test" {
+  brand_id = tolist(data.okta_brands.test.brands)[0].id
+}
--- a/examples/resources/okta_customized_signin_page/basic.tf
+++ b/examples/resources/okta_customized_signin_page/basic.tf
@@ -1,0 +1,14 @@
+resource "okta_brand" "test" {
+  name   = "testBrand"
+  locale = "en"
+}
+
+resource "okta_customized_signin_page" "test" {
+  brand_id       = resource.okta_brand.test.id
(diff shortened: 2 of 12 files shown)


tgoodsell-tempus:
@monde @duytiennguyen-okta Just wanted to ping you for feedback / review on this. I wasn't super sure how to write into the "test code", but the check I've done capturing the local test requests is ensuring that the UPDATE `PUT` call does not include the `client_secret` value in the request. 

If the recorded tests for this demonstrate that, we should be good to go.

Additionally, would it be possible for you to clarify in the `CONRTRIBUTING.MD` if you would prefer that we record & push our `go-vcr` tests along with our PRs here? Or do you prefer to record all of them using some standard instances you manage?

duytiennguyen-okta:
Thanks @tgoodsell-tempus. We will review it and let you know

monde:
@tgoodsell-tempus we're going to bring in this PR. It's a little bit more and than just a bug fix and we'd like a CLA. I don't think we have a CLA on file for you https://developer.okta.com/cla/
Ping me at mike.mondragon@okta.com and we can discuss further.

monde:
Okta internal reference https://oktainc.atlassian.net/browse/OKTA-671286

tgoodsell-tempus:
FYI @monde, I submitted the CLA to Okta earlier today. Hopefully y'all get the confirmation on your end soon enough.

tgoodsell-tempus:
FYI @monde @duytiennguyen-okta This moved to https://github.com/okta/terraform-provider-okta/pull/1852 due to a rebase gone wrong.

PR #1852, opened 2023-12-16
Okta App OAuth: Make Omit Secret Safer To Set

tgoodsell-tempus:
The goal for this PR was:

- Ensure that when we actually do the UPDATE where omit_secret is activated, that update does not send the client_secret value stored in state to the Okta API.
- Add comments and fix the state setting, to better reflect the reality of what this function is doing
- We also added "no-op" but semantic changes to how the return data from the app calls is used. Basically we use the output value instead of the passed in request, even though this is the same since the lower level API calls use that object to fill in the data from the response. It makes more "sense" from the Golang SDK standpoint to specifically show we're using the results of the API calls to set the state values, instead of the more ambiguous method currently in use.
- Update document to explain more clearly the functions of the omit_secret and client_basic_secret arguments, as well as the client_secret attribute.

Originally was: https://github.com/okta/terraform-provider-okta/pull/1815

However a merge/rebase gone wrong murdered the git history which required more significant repair. GH closed my other PR due to this.


tgoodsell-tempus:
@monde @duytiennguyen-okta This should be rebased correctly and ready for testing and validation on your end.

PR #1894, opened 2024-02-08
Handle multiple apps returned by SAML app data source, where one app exactly matches.

MatthewJohn:
In the case where there are two apps: "MyApp" and "MyApp 2".
If the data source for saml apps queried with label "MyApp" and the API returns "MyApp 2" as the first app in the list, the data source will fail.
This change now iterates through all of the apps returned and returning an exact match, if "label" has been specified.

### DIFF
--- a/okta/data_source_okta_app_saml.go
+++ b/okta/data_source_okta_app_saml.go
@@ -326,6 +326,7 @@ func dataSourceAppSamlRead(ctx context.Context, d *schema.ResourceData, m interf
 				if appItx.Label == filters.Label {
 					app = appItx
 					foundMatch = true
+					break
 				}
 			}
 			if !foundMatch {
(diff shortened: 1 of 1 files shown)


duytiennguyen-okta:
@MatthewJohn Thanks. We will review it

MatthewJohn:
Thanks @duytiennguyen-okta :)

okta/data_source_okta_app_saml.go
```diff
@@ -319,11 +319,22 @@ func dataSourceAppSamlRead(ctx context.Context, d *schema.ResourceData, m interf
...
 		}
-		if filters.Label != "" && appList[0].Label != filters.Label {
-			return diag.Errorf("no SAML application found with the provided label: %s", filters.Label)
+
+		if filters.Label != "" {
+			foundMatch := false
+			for _, appItx := range appList {
+				if appItx.Label == filters.Label {
+					app = appItx
+					foundMatch = true
```
duytiennguyen-okta, line 328: It should break here, similar to https://github.com/okta/terraform-provider-okta/blob/master/okta/data_source_okta_app.go#L111?

duytiennguyen-okta (changes requested):

duytiennguyen-okta:
If you're ok with the suggestion, I will bring the PR in and update it @MatthewJohn

okta/data_source_okta_app_saml.go
MatthewJohn: Oh course, yes, apologies!

PR #1898, opened 2024-02-12
update brand to correctly reflect api

duytiennguyen-okta:


mvitt:
I believe you would have to remove the "brand_id" attribute from the "okta_email_domain" resource if you add the above cause if not you will re-introduce the cyclical error we saw previously.

Also, having the "brand_id" on the "okta_email_domain" resource will prevent that "okta_email_domain" from being used by multiple brands

duytiennguyen-okta:
@mvitt Email domain required brand_id to be created. It is the way the API was designed

mvitt:
@duytiennguyen-okta If I would like to use the same custom email domain for all my brands, how would I go about doing that?

Say that I have five brands and I would like to use the same email domain for all of them, it would make sense to have only one "okta_email_domain" that is added to each "okta_brand" instead of have five "okta_email_domain" resources with the same values except for the brand_id. Especially since "okta_email_domain" spits out a new "dns_validation_records" which must be passed to aws route53, etc (This would cause errors in route53, right)

duytiennguyen-okta:
@mvitt you would have to update it through okta_brand. I know it is awkward, but it is the way the API is designed.

mvitt:
@duytiennguyen-okta as Reece points out, there seems to be a conflict between brands and email domain. Is there a way you could pull in the API team to review these two APIs?

duytiennguyen-okta:
Actually this isn't going to work. Close the PR now

duytiennguyen-okta:
@mvitt yeah I will have to do that. cc @jefftaylor-okta

PR #1910, opened 2024-02-23
Matthew john 1894

duytiennguyen-okta:
Close #1894

### DIFF
--- a/okta/data_source_okta_app_saml.go
+++ b/okta/data_source_okta_app_saml.go
@@ -326,6 +326,7 @@ func dataSourceAppSamlRead(ctx context.Context, d *schema.ResourceData, m interf
 				if appItx.Label == filters.Label {
 					app = appItx
 					foundMatch = true
+					break
 				}
 			}
 			if !foundMatch {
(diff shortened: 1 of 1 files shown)


monde (approved):

okta/data_source_okta_app_saml.go
```diff
@@ -319,11 +319,25 @@ func dataSourceAppSamlRead(ctx context.Context, d *schema.ResourceData, m interf
...
 		}
-		if filters.Label != "" && appList[0].Label != filters.Label {
-			return diag.Errorf("no SAML application found with the provided label: %s", filters.Label)
+
+		// Okta API for list apps uses a starts with query on label and name
+		// which could result in multiple matches on the data source's "label"
+		// property.  We need to further inspect for an exact label match.
+		if filters.Label != "" {
+			// guard on nils, an app is always set
+			app = appList[0]
```
MatthewJohn, line 328: @duytiennguyen-okta I see you've modified to perform initial assignment, but this will assign it to the first app, even if it doesn't match and there's no logic to fail if a matching app isn't found in the for-loop

MatthewJohn: Is it important that app is always set? Given that if it isn't, we were returning a diag error in https://github.com/okta/terraform-provider-okta/pull/1910/commits/912cfb5fcf4790498b3a0e69d433dfb5b0e0c04f#diff-bd713a1119c4af963b9c842b006fdbe0a563da383f7172d61e764375ce3831e6L326

PR #1967, opened 2024-04-15
Feat: DataSource okta_apps

exitcode0:
I've cherry picked an existing PR and had a crack at implementing this with the terraform provider framework
I'm still pretty clueless with GoLang and would love feedback so I can tweak this as needed

I had some trouble trying to get additional information to work with the dataSource, things like the appSettings, appVisibility, appCredentials - I couldn't figure out how to go about creating an interface for these types, i'd love some guidance on this if possible

### DIFF
--- a/okta/app_filter.go
+++ b/okta/app_filter.go
@@ -67,3 +67,4 @@ func getAppFilters(d *schema.ResourceData) (*appFilters, error) {
 	}
 	return filters, nil
 }
+
--- a/okta/data_source_okta_apps.go
+++ b/okta/data_source_okta_apps.go
@@ -1,0 +1,220 @@
+package okta
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/path"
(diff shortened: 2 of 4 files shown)


exitcode0:
@duytiennguyen-okta any chance I could trouble you for a code review on this one?
i'd love some guidance on how I should've gone about constructing interfaces for these applications or if i've gone about this in the wrong way

duytiennguyen-okta:
@exitcode0 hey nice works. The apps seems to works fine and got all the fields you need it seems. Before approval, I just want to know if there is no parameter provider, `okta_apps` will return all apps? Is that your intention?
`data "okta_apps" "test" {
}
`
Also do you want to bring appSettings, appVisibility and appCredentials in still? I can give you an example if you wish

exitcode0:
Hi @duytiennguyen-okta

> if there is no parameter provider, okta_apps will return all apps? Is that your intention?

That would be my intention, I believe this is how the `data "okta_groups" "test" { }` works today

> Also do you want to bring appSettings, appVisibility and appCredentials in still? I can give you an example if you wish

I'm open to it but hadn't attempted it as I was having had some troubles getting things to work properly with the GoLang SDK, I think this is due to my lack of GoLang experience, I'm probably missing something obvious

I was having troubles handling the response from `GetActualInstance()` as this could return a list containing multiple Types, with each of those types containing a different type implementation for something like `appSettings`, e.g. `OpenIdConnectApplicationSettings` & `SamlApplicationSettings`

I wasn't sure if I should be creating my own interface here, or perhaps if I should have used a more generic type or interface from the GoLang SDK

duytiennguyen-okta:
https://github.com/okta/terraform-provider-okta/pull/2136. This is what I did for app visibility. For you question, I don't think that is possible to handle multiple type implementation for appSettings using generics or interface

PR #1993, opened 2024-05-17
Fix okta_app_* unable to import authentication policy

### LINKED ISSUE #1742: UnPlanned changes - `okta_app*` - unexpected `authentication_policy`
### Community Note

### Affected Resource(s)
- okta_app*

```hcl
resource "okta_app_bookmark" "example" {
  label  = "Example"
  url    = "https://example.com"
}
```

### Expected Behavior

`Terraform Plan` should outline all changes that would result from executing `Terraform Apply` 

### Actual Behavior

When the upstream API chooses default behavior we don't expose this in the `Terraform Plan`

In this case the existing well known behavior is that omitting `authentication_policy` on `okta_app*` resources results in the default authentication policy being used
This can unexpectedly replace a authentication policy that was click-ops provisioned to an application

I believe this has something to do with this API not supporting partial updates and assuming that an omitted value is equivalent to a null value

### Steps to Reproduce

1. `terraform apply`
2. Click-Ops change Auth Policy
3. `terraform apply`
4. Auth Policy has now changed in a way that was not visible to IAC

### Important Factoids

Most of our engineers expect that `Terraform Plan` will outline all changes that would result from executing `Terraform Apply` 
When this contract is broken, it understandably catches most engineers unaware and can cause outages

One of the benefits of Terraform is that it can be rolled out gradually because it does not attempt to modify resources that are not under management
But in this case this is being undermined by default behaviors in the upstream API

Some ideas below for paths forward to more intuitive behavior
Things within Okta Terraform's control
 - Surface what will happen in the `Terraform plan` by manually constructing plan info (not sure if this is possible)
 - Require the attribute on `okta_app*` resources
   - support `"default"` as a string
   - require using the ID from a Data Source

Things outside the Okta TF provider's control
 - Raise an issue upstream team that owns the API
   - support partial updates on the API in question (all fields)
   - leave authentication policy unchanged if it was omitted
   - question why omitted is treated as equivalent to `null` (requiring explicit null-ish values on endpoints that don't support partial updates seems like less of a footgun imo)

### LINKED ISSUE #1942: okta_app_bookmark import missing authentication_policy
### Community Note

### Terraform Version

```shell
% terraform -v
Terraform v1.7.5
on darwin_arm64
```

### Affected Resource(s)

- okta_app_bookmark

```hcl
terraform {
  backend "local" {
    path = "terraform.tfstate"
  }
  required_providers {
    okta = {
      source = "okta/okta"
      version = "4.8.0"
    }
  }
}

provider "okta" {}

resource "okta_app_bookmark" "example" {
  label  = "Terraform Test"
  url    = "https://example.com"
  authentication_policy = "12345"
}
```

### Debug Output

```shell
terraform init
terraform plan
# Shows it is going to add the app which is correct (as it has not been imported yet)
terraform import okta_app_bookmark.example abcde
terraform plan
# Now shows it is going to set the authentication policy which should have been imported
```

Plan Output
```shell
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following
symbols:
  ~ update in-place

Terraform will perform the following actions:

  # okta_app_bookmark.example will be updated in-place
  ~ resource "okta_app_bookmark" "example" {
      + authentication_policy      = "12345"
        id                         = "abcde"
        name                       = "bookmark"
        # (11 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.

```

### Panic Output
n.a.

### Expected Behavior
Terraform app was imported so it should not be making any changes to the authentication policy. In the Okta Admin the authentication_policy is set to the same value as in the Terraform configuration so no changes should be on the Terraform plan output (after import).

n.a.

n.a.

### Actual Behavior

(see above)

### Steps to Reproduce

(see above)

### Important Factoids

n.a.

### References

* I have tried to raise this with Okta Support Case: 02017270

### Notes

* I suspect this is the same for other applications (but I have not checked)

duytiennguyen-okta:
Close #1942 
Close #1742

### DIFF
--- a/okta/resource_okta_app_bookmark.go
+++ b/okta/resource_okta_app_bookmark.go
@@ -40,6 +40,7 @@ other arguments that changed will be applied.`,
 			"authentication_policy": {
 				Type:        schema.TypeString,
 				Optional:    true,
+				Computed:    true,
 				Description: `The ID of the associated app_signon_policy. If this property is removed from the application the default sign-on-policy will be associated with this application.`,
 			},
 		}),
@@ -82,7 +83,7 @@ func resourceAppBookmarkRead(ctx context.Context, d *schema.ResourceData, m inte
 		d.SetId("")
 		return nil
 	}
-	setAuthenticationPolicy(d, app.Links)
+	setAuthenticationPolicy(ctx, m, d, app.Links)
 	_ = d.Set("url", app.Settings.App.Url)
 	_ = d.Set("request_integration", app.Settings.App.RequestIntegration)
 	appRead(d, app.Name, app.Status, app.SignOnMode, app.Label, app.Accessibility, app.Visibility, app.Settings.Notes)
--- a/okta/app_authentication_policy.go
+++ b/okta/app_authentication_policy.go
@@ -17,14 +17,14 @@ func createOrUpdateAuthenticationPolicy(ctx context.Context, d *schema.ResourceD
 	return err
 }
 
-func setAuthenticationPolicy(d *schema.ResourceData, links interface{}) {
-	// only set the authentication_policy if it's been set previously so we can
-	// get proper change detection on the value
-	if _, ok := d.GetOk("authentication_policy"); ok {
-		accessPolicy := linksValue(links, "accessPolicy", "href")
-		if accessPolicy != "" {
-			d.Set("authentication_policy", path.Base(accessPolicy))
-		}
+func setAuthenticationPolicy(ctx context.Context, m interface{}, d *schema.ResourceData, links interface{}) {
+	// setAuthenticationPolicy by default, switching from optional to optional and computed
+	if isClassicOrg(ctx, m) {
+		return
+	}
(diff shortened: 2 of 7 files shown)


okta/resource_okta_app_bookmark.go
```diff
@@ -40,6 +40,7 @@ other arguments that changed will be applied.`,
 			"authentication_policy": {
 				Type:        schema.TypeString,
 				Optional:    true,
+				Computed:    true,
```
exitcode0, line 43: I don't have confidence in my knowledge of TF provider schemas and their behaviours

Will setting `Computed: true` here prevent the user from specifying or updating the  authentication policy on a given `okta_app_*`?

duytiennguyen-okta: No. It will allow the provider to set the authentication_policy to default if it is not specified. I want to deprecate authentication_policy, but it will have to be major version change

monde (approved):

PR #1997, opened 2024-05-23
fix crash when creating resource set from devices

duytiennguyen-okta:
There is no links to self when create resource set from devices, causes the crash

```
"_links": {
   "self": null
}
```

### DIFF
--- a/okta/resource_okta_resource_set.go
+++ b/okta/resource_okta_resource_set.go
@@ -142,9 +142,12 @@ func flattenResourceSetResources(resources []*sdk.ResourceSetResource) *schema.S
 			links := res.Links.(map[string]interface{})
 			var url string
 			for _, v := range links {
-				for _, link := range v.(map[string]interface{}) {
-					url = link.(string)
-					break
+				_, ok := v.(map[string]interface{})
+				if ok {
+					for _, link := range v.(map[string]interface{}) {
+						url = link.(string)
+						break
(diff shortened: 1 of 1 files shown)


okta/resource_okta_resource_set.go
```diff
@@ -142,14 +142,21 @@ func flattenResourceSetResources(resources []*sdk.ResourceSetResource) *schema.S
...
 			var url string
 			for _, v := range links {
-				for _, link := range v.(map[string]interface{}) {
-					url = link.(string)
-					break
+				_, ok := v.(map[string]interface{})
+				if ok {
+					for _, link := range v.(map[string]interface{}) {
+						url = link.(string)
+						fmt.Println("149", url, res.Id)
```
monde: cruft?

okta/resource_okta_resource_set.go
```diff
@@ -142,14 +142,21 @@ func flattenResourceSetResources(resources []*sdk.ResourceSetResource) *schema.S
...
+						break
+					}
+				} else {
+					fmt.Println("152", res.Id)
 				}
 			}
 			arr = append(arr, url)
 		}
 	}
+	fmt.Println("159", arr)
```
monde: cruft

monde (changes requested):

monde (approved):

PR #2001, opened 2024-05-31
update doc

duytiennguyen-okta:


monde (approved):

tyler-hc:
was this change related to any issues? I just noticed it was added recently and I'm trying to find what problems using for_each could cause. I'm using it for a few things, but was using them before this document change. It'd be nice to know the problem and if we should work on removing existing resource_okta_app_group_assignments that are using for_each.

PR #2032, opened 2024-07-02
revert d.GetOk back to d.Get for resourceProfileMapping

zztczcx:
This PR https://github.com/okta/terraform-provider-okta/pull/2004/files 
added importing, but also changed the logic around `delete_when_absent`. 

if `delete_when_absent` was set to `false`
deleteWhenAbsent, ok := d.GetOk("delete_when_absent")

deleteWhenAbsent is false,
ok is false too.  it's because the `ok` means whether or not the key has been set to a non-zero value. 
However boolean's zero value is false. 

so the new logic introduced by PR 2004 breaks the old logic.

### DIFF
--- a/okta/resource_okta_profile_mapping.go
+++ b/okta/resource_okta_profile_mapping.go
@@ -111,8 +111,7 @@ func resourceProfileMappingCreate(ctx context.Context, d *schema.ResourceData, m
 	}
 	d.SetId(mapping.Id)
 	newMapping := buildMapping(d)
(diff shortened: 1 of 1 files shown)


jz-latitude:
@duytiennguyen-okta  please review this PR when you are free. thanks.

duytiennguyen-okta:
Yeah I don't understand why ok is false when deleteWhenAbsent is false

duytiennguyen-okta:
Close. Refer to #2053

PR #2053, opened 2024-07-31
revert import profile mapping

duytiennguyen-okta:
Revert due to weird behaviour around [GetOK()](https://github.com/okta/terraform-provider-okta/pull/2032)


monde (approved):

pro4tlzz:
Thanks, the recent change to this resource caused our configuration to state all profile mapping resources would be destroyed

Issue #1573, opened 2023-05-26
`inactivity_period` in `okta_app_signon_policy_rule` should not default to `PT1H`


chickenriggies:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

Terraform v1.4.5
on darwin_amd64
+ provider registry.terraform.io/okta/okta v4.0.0

### Affected Resource(s)

- okta_app_signon_policy_rule

### Terraform Configuration Files

```hcl
resource "okta_app_signon_policy_rule" "fido" {
  name                                   = "fido"
  policy_id                              = okta_app_signon_policy.fido.id
  access                                 = "ALLOW"
  factor_mode                            = "2FA"
  groups_included                        = [
    local.groups["Linux Users"].id
  ]
  network_connection                     = "ANYWHERE"
  priority                               = 98
  re_authentication_frequency            = "PT43800H"
  constraints                            = [
    jsonencode({
      "possession" : {
        "types" : [
          "security_key"
        ],
        "methods" : [
          "webauthn"
        ]
      }
    })
  ]
}
```

### Debug Output

https://gist.github.com/hb3b/14259a23846b15ce848288e832845bdd

### Expected Behavior

`inactivity_period` should not be set unless deliberately set by the user.

#### Can this be done in the Admin UI?

`inactivity_period` does not appear to be configurable in the Admin UI.

#### Can this be done in the actual API call?

Yes, through omission.

### Actual Behavior

`inactivity_period` is set to a default value of: `PT1H`.

### References

- https://github.com/okta/terraform-provider-okta/pull/1184
- https://registry.terraform.io/providers/okta/okta/latest/docs/resources/app_signon_policy_rule#inactivity_period

cjh-Ella:
Hello,

we encountered the same issue. Can we reopen this issue?

Thank you!

pro4tlzz:
For now you can ignore this specific attribute, I too ran into this issue

duytiennguyen-okta:
https://oktainc.atlassian.net/browse/OKTA-728926

jbow-square:
[...]

PR #2070, opened 2024-08-20
Remove inactivity_period default value from app_signon_policy_rule

reinoudk:
This addresses #1573 which, even though closed as `stale`, I think should be fixed. 

For example, when creating a rule through the UI or when not explicitly setting this value through the API, the JSON that is returned when subsequently pulling the rule from the API will also not have a value for `inactivityPeriod`. The provider docs do not specify a default value either.

Note: I realize this could considered be a breaking change w.r.t. how the provider handled this value up until now 😬.


duytiennguyen-okta:
@reinoudk yes, it is a breaking change and will have to wait for v5 to be released

exitcode0:
Should this PR be rebased now that `V5` has been released?
We could merge this into a long running branch for `V6` if we want?

PR #2105, opened 2024-10-18
add rlsecure

duytiennguyen-okta:


bretterer (changes requested):
I think `go mod vendor` will be what we need here instead of just `install`.... we need to have the vendor files in the source code for the scan

bretterer (dismissed):

monde (dismissed):

monde (approved):

PR #2153, opened 2024-12-04
add support for catch_all = deny

### LINKED ISSUE #1251: Add attribute to okta_app_signon_policy to create a 'Catch All' rule that's 'DENY'
okta_app_signon_policy creates a default 'Catch All' rule.  Once such a rule is created, it's difficult to change anything about it with terraform since the codebase implementing okta_app_signon_policy_rule is designed to work with mutable rules vs. all the restrictions associated with the 'Catch All' rule.

If you could add an attribute to okta_app_signon_policy (catchAll = "DENY"), then okta_app_signon_policy could create policies with 'Catch All' being a noop - the mutable rules could be used for everything else.  As it stands, I have to go and manually configure the 'Catch All' rules to 'DENY' while using terraform scripts to configure all the other rules.

duytiennguyen-okta:
Close #1251


monde (approved):
:shipit:

PR #2187, opened 2025-01-24
fix issue relate to os_expression does not distinguished between null…

duytiennguyen-okta:
… and empty string


monde (approved):
:shipit:

PR #2207, opened 2025-02-05
Improve `app_oauth` docs for "wildcard_redirect"

exitcode0:
This is a non-exhaustive list that includes values i've found to be valid from my testing and reversing the Okta Admin Console
It's possible there are other valid values that i'm not aware of, hence not adding a validator to the schema here


monde (approved):
Thanks @exitcode0

Issue #2208, opened 2025-02-06
Create `V5` branch for community contributions


exitcode0:
A [recent comment](https://github.com/okta/terraform-provider-okta/issues/1183#issuecomment-2626353881) has made me uncertain about Okta's future support plans for the Okta Terraform provider, whether that's `V4` or `V5` and beyond

While I'm sure more details will be shared about this soon,
I'd like to ask that one of the maintainers to create a "V5" branch in this repo.

This will give the community a place to merge PRs that are to disruptive for `V4`/`master` and allow the community to contribute incremental progress towards a `V5` release.
Doing this will also prevent these more disruptive PRs from becoming stale and accumulating merge conflicts or un-addressed `TODO`s

monde:
We're in the process of standing up a new team to take ownership of the Okta Terraform Provider. See [v5.0.0 plan](https://github.com/okta/terraform-provider-okta/issues/2230). It will be up to that new team how to manage community contributions. Until then we'll continue taking community PRs as our bandwidth permits and/or is elevated internally through a customer support priority escalation process.

PR #2210, opened 2025-02-10
chore: refactor `data_app_group` to use plugin-framework

exitcode0:
I've not yet tested this code

Also not sure if the test cases will need to be re-written for `terraform-plugin-framework`

### DIFF
--- a/okta/services/idaas/data_source_okta_app_group_assignments.go
+++ b/okta/services/idaas/data_source_okta_app_group_assignments.go
@@ -2,50 +2,81 @@ package idaas
 
 import (
 	"context"
+	"fmt"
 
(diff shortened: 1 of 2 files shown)


monde:
Thanks @exitcode0 
Is there a bug here? I'm trying to harden up the reliability of the Okta TF releases and I don't want to take up code changes in behavior implementation unless they are are fixing a bug.

exitcode0:
Hey @monde,

No bug here, this is just a general refactor
My understanding is that we'd want to move to plugin-framework for a `v5`?
If we action #2208 I think it makes more sense for this PR target that branch instead of `master`

PR #2211, opened 2025-02-10
Improve `app_oauth` docs for "token_endpoint_auth_method"

exitcode0:
Update description for "token_endpoint_auth_method" to include a non-exhaustive list of values that are **_**currently_** valid

Values may be added or removed in the future which is why I didn't add a schema validator here


monde (approved):
Thanks @exitcode0

PR #2217, opened 2025-02-12
Improve `limit` implementation on data_okta_groups

exitcode0:
We were previously only allowing for the use of a higher `limit` if the `q` parameter was used
This new implementation should allow the user to override the default page size when they need to
This should also improve the response time for queries that were seeking a small number of groups using the `q` parameter as the okta api returns results faster when the default page size is used

### DIFF
--- a/docs/data-sources/groups.md
+++ b/docs/data-sources/groups.md
@@ -21,6 +21,7 @@ data "okta_groups" "example" {
 
 ### Optional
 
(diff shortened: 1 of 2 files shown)


exitcode0:
Hey @monde, @pranav-okta, @aditya-okta 
Could I please get a review on this PR?
It'd be great if I can make use of an increased `limit` on data sources that aren't using the `q` query parameter

PR #2218, opened 2025-02-12
Add schema validators for app notes

exitcode0:
I think it is unlikely that the max_length of admin notes will change frequently or soon, thus IMO it should be safe for us to do input validation here.


exitcode0:
@pranav-okta - Could I please get a review on this one
it should reuse the same validator function added in #2219

pranav-okta (approved):
👍

PR #2219, opened 2025-02-12
add schema validators for name length - `okta_group` & `okta_group_rule`

exitcode0:
Requests to increase the max length of these fields now date back to at least 7 years ago

- https://ideas.okta.com/app/#/case/204022 (6 months ago)
- https://ideas.okta.com/app/#/case/106505 (6 years ago)
- https://ideas.okta.com/app/#/case/107116 (7 years ago)

Given the above, I think it is unlikely that this behavior will change frequently or soon, Therefore, it should be safe for us to do input validation here IMO


exitcode0:
@pranav-okta Could I trouble you for a review on this PR please? 🤞

pranav-okta (approved):
🚀

PR #2229, opened 2025-02-18
feat: smart card idp policy support

adenix:
added smart_card_idp as an authenticator provider option


duytiennguyen-okta:
Close in favor of #2294

PR #2233, opened 2025-02-24
Correct `.gitignore` for .env files

exitcode0:
`.env` seemed to only ignore files called exactly `.env` in my IDE 
`*.env` ignored all env files at the top level of the repo
`**.env` ignored env files located at any level throughout the repo and thus seemed the most appropriate to me


monde (approved):
Thanks @exitcode0

PR #2253, opened 2025-03-20
docs: add mfa_lifetime usecase

pranav-okta:
The Terraform [documentation](https://registry.terraform.io/providers/okta/okta/latest/docs/resources/policy_rule_signon) for the `okta_policy_rule_signon` does not specify anything about the `mfa_lifetime`.  The `mfa_lifetime` is only valid when the `mfa_prompt` is either `ALWAYS` or `SESSION`. Since this is not specified anywhere, there are issues where the `mfa_lifetime` is used even if the `mfa_prompt` is invalid. This PR adds docs about the specification.


aditya-okta (approved):
LGTM

PR #2258, opened 2025-03-28
fix: make HTTP 423 retryable for `okta_group_rule`

### LINKED ISSUE #2151: Policy lock exception, HTTP Status: 423 Locked
Hello, 

We are having hundreds of Okta group rules which are updated simultaneously. We first create a group and then a group rule where the group is being used in the rule policy. The code works fine but sometimes some of the group rule requests fails with the below error, while creating the resource. 

E0000239: Policy lock exception
HTTP Status: 423 Locked
Policy priorities are being reconciled. Please try again later.

We discussed the problem with Okta and they mentioned that "the SDK does not have parallelism built into it". As Terraform is running with 10 parallel threads by default, it is ending up into this problem. Okta suggests that this should be handled at the client end. In this case it is the Okta Terraform provider which we are using. 

The error doesn't appear anymore when we run apply with the flag -parallelism=1. However, this increases the amount of time the apply runs. 

Could you please look into the issue and provide us a fix for the same?

Thanks in advance

### Terraform Version

Terraform v1.9.5

### Affected Resource(s)

- okta_group_rule

### Expected Behavior

No error

I don't know 

I don't know

### Steps to Reproduce

1. Create 100+ groups
2. Create 100+ group rules, use the groups from the above in the group rule policy
3. `terraform apply`

exitcode0:
fixes #2151

<hr>
I've reached out to Okta support about this behavior in Support Cases `02157630` and `02192142`

I was told that there are no plans to change the HTTP `423` response code
The HTTP `423` response code is sometimes returned by the Okta API when modifying multiple group rules in rapid succession.  
The current behavior when the provider gets a HTTP `423` is to bail out from the entire `terraform apply`  
This is an over reaction (imo) as the request will often succeed if immediately retried

<hr>

I'm open to suggestions and/or guidance about how to implement this change, especially if there is a better approach I should take instead

I added the logic to the `okta_group_rule` resource as I wanted to limit the change in behavior/functionality to only this resource, which rules out making the change to the SDK request executor(s)

### DIFF
--- a/examples/resources/okta_group_rule/test_423_response_capture.tf
+++ b/examples/resources/okta_group_rule/test_423_response_capture.tf
@@ -1,0 +1,53 @@
+resource "okta_group" "test" {
+  name = "test_423_response_capture"
+}
+
+# Main group rule that should succeed
+resource "okta_group_rule" "test" {
+  name              = "test_423_response_capture"
+  status            = "ACTIVE"
+  group_assignments = [okta_group.test.id]
+  expression_type   = "urn:okta:expression:1.0"
+  expression_value  = "String.startsWith(user.firstName,\"andy\")"
+}
+
+# Additional rules that might trigger 423s but won't fail the test
+resource "okta_group_rule" "concurrent1" {
+  name              = "test_423_response_capture_concurrent1"
+  status            = "ACTIVE"
+  group_assignments = [okta_group.test.id]
+  expression_type   = "urn:okta:expression:1.0"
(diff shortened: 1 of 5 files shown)


duytiennguyen-okta (dismissed):

emtunc:
hey team, apologies for the ping @duytiennguyen-okta - can we expect this to go into the provider in the coming weeks?

pranav-okta (approved):

Issue #2160, opened 2024-12-10
Cannot modify the default authentication policy


MScottBlake:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

1.9.8
1.10.1

### Affected Resource(s)

- okta_app_signon_policy

### Terraform Configuration Files

```hcl
resource "okta_app_signon_policy" "this" {
  name        = var.name
  description = var.description

  lifecycle {
    create_before_destroy = true
  }
}
```

### Debug Output

On change, I might get something similar to this when I run `terraform plan`.
```
Terraform used the selected providers to generate the following execution
plan. Resource actions are indicated with the following symbols:
  ~ update in-place
Terraform will perform the following actions:
  # module.standard_policy.okta_app_signon_policy.this will be updated in-place
  ~ resource "okta_app_signon_policy" "this" {
      ~ description = "Our standard authentication policy. Users must use a managed device on macOS/Windows and a registered device on Android/iOS. Managed by Terraform at https://examplle.com/okta/authentication-policies." -> "Our standard authentication policy. Users must use a managed device on macOS/Windows and a registered device on Android/iOS. Managed by Terraform at https://example.com/iam/okta-authentication-policies."
        id          = "********************"
        name        = "Standard Policy"
    }
Plan: 0 to add, 1 to change, 0 to destroy.
```

### Expected Behavior

The authentication policy should have a new description.

#### Can this be done in the Admin UI?

Yes

#### Can this be done in the actual API call?

I don't know

### Actual Behavior

```
module.standard_policy.okta_app_signon_policy.this: Modifying... [id=********************]
╷
│ Error: failed to update authentication policy: the API returned an error: Cannot modify the priority attribute because it is read-only.
│ 
│   with module.standard_policy.okta_app_signon_policy.this,
│   on okta_authentication_policy/main.tf line 1, in resource "okta_app_signon_policy" "this":
│    1: resource "okta_app_signon_policy" "this" {
```

### Steps to Reproduce

1. Import the default authentication policy.
2. Modify a value
3. `terraform apply`

[...]

PR #2259, opened 2025-03-31
Okta 899300

### LINKED ISSUE #2246: Cannot Upgrade to 4.15.0 with Default Authentication Policy
### Community Note

### Terraform Version

```
Terraform v1.11.1
on darwin_arm64
+ provider registry.terraform.io/okta/okta v4.15.0
```

### Affected Resource(s)

- `okta_app_signon_policy`

```hcl
terraform {
  required_providers {
    okta = {
      source  = "okta/okta"
      version = "4.15.0"
    }
  }
}

resource "okta_app_signon_policy" "default" {
  name        = "Default Policy"
  description = "<CURRENT DESCRIPTION FOR DEFAULT AUTHENTICATION POLICY>"
}

import {
  id = "<ID FOR DEFAULT AUTHENTICATION POLICY>"
  to = okta_app_signon_policy.default
}
```

### Debug Output

(Did not attach due to personal information included; please advise as to how to send the file securely if required.)

### Panic Output

N/A

### Expected Behavior

The Terraform apply for upgrading the provider should succeed.

N/A

N/A (the API call failed here, but the API call succeeds with valid input)

### Actual Behavior

Because the `catch_all` attribute is added, but it is not populated when the state is read from the server, Terraform detects a change in the field from `null` to `true`.

Therefore, it triggers the `Update` function for authentication policy in the provider. The provider makes a call to `PUT /api/v1/policies/<POLICY_ID>` to update the default authentication policy. Here, the body of the request has no `priority` set.

For non-default policies, `priority` set to null works in update. However, it seems that the API has special validation logic for the default policy, and rejects unset / `null` in the `priority` field.

(Testing outside of the provider, the API to update the policy works when the `priority` is set to `1`, the current value.)

### Steps to Reproduce

1. `terraform apply`

### Important Factoids

N/A

### References

- #2160 is a related issue

pranav-okta:
The Default Authentication Policy can be modified through the API but only a few attributes. i.e, `description`. Our provider is not adding the `policy` which is not in the state. As a result, it tries to update the policy to a null value. 
This PR adds `priority` and sets it to the default value of 1. Please see [this](https://developer.okta.com/docs/api/openapi/okta-management/management/tag/Policy/#tag/Policy/operation/replacePolicy) API doc.

Closes #2246 and #2160

### DIFF
--- a/docs/resources/app_signon_policy.md
+++ b/docs/resources/app_signon_policy.md
@@ -95,6 +95,7 @@ resource "okta_app_signon_policy_rule" "some_rule" {
 ### Optional
 
 - `catch_all` (Boolean, default true, creation-only argument) If false, the default rule of the policy is set access to `DENY`. Otherwise default behavior of the default rule is to leave access at `ALLOW`.  **WARNING** setting this attribute to false changes policy rule's default behavior. Use at your own risk. This is only applied during creation and does not affect import or update.
+- `priority` (Default 1) Specifies the order in which this policy is evaluated in relation to the other policies
 
 ### Read-Only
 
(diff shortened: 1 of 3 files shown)


monde (approved):
:shipit: 🎉

PR #2282, opened 2025-04-16
add support for type AUTH_METHOD_CHAIN

duytiennguyen-okta:
add support for type AUTH_METHOD_CHAIN for resource `okta_app_signon_policy_rule`

### DIFF
--- a/sdk/v2_verificationMethod.go
+++ b/sdk/v2_verificationMethod.go
@@ -1,4 +1,3 @@
-// DO NOT EDIT LOCAL SDK - USE v3 okta-sdk-golang FOR API CALLS THAT DO NOT EXIST IN LOCAL SDK
 package sdk
 
 type VerificationMethod struct {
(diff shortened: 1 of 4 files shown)


monde (approved):

sdk/v2_verificationMethod.go
```diff
@@ -1,4 +1,3 @@
-// DO NOT EDIT LOCAL SDK - USE v3 okta-sdk-golang FOR API CALLS THAT DO NOT EXIST IN LOCAL SDK
```
exitcode0, line 1: 😅

delize: > It's a bold strategy, Cotton. Let's see if it pays off for 'em

But seriously, this doesn't seem like a good idea? Would it be better to add a supplemental verificationMethod file that is called? I don't know the in's and out's of Go, but am 100% onboard with this getting merged in, just feels like this could be a breaking change?

exitcode0: TL;DR: I think they're making changes to this file so they can expedite support for this feature ❤️ 

The old SDK is littered with these "do not edit" comments as the maintainers want to move away from maintaining a fork of the SDK
This is a good thing IMO as we should be upstreaming any patches rather than maintaining a bundled fork of the SDK

that said, these changes are pretty harmless IMO, they're just adding the missing type declarations to the old SDK so that it can support the new feature
I think the author did this to expedite support for this feature for you

PR #2320, opened 2025-05-21
fix: HTTP 404 in app_oauth_role_assignment

exitcode0:
if we get a 404 its because the resource doesn't exist when this happens we should remove the resource from state to prevent a deadlock where a manually replaced resource can't be deleted because the plan fails due to it looking up a now defunct ID


exitcode0:
Hey @aditya-okta or @pranav-okta,
Could I please ask you to approve the Github Workflow run for this PR so the Github Checks can run? 🙏

PR #2323, opened 2025-05-26
fix: HTTP 404 handling okta_app_group_assignments

exitcode0:
The function `deleteGroupAssignments ` trigger an error if we get a 404 response when we try to remove an `app_group_assignment`

If we get a 404 in this circumstance, it can only mean one of the following:
- The App ID is wrong
- The Group ID is wrong
- The Group no longer exists
- The App no longer exists
- The Group is no longer assigned to the app

In all of these cases we don't need to throw an error as the group is no longer assigned to the application in all cases
Therefore, this function should suppress the a 404 response if it gets one, this aligns with functions elsewhere in the file that also call `suppressErrorOn404`


exitcode0:
Hey @aditya-okta,
Could I trouble you for a review on this PR?

aditya-okta (approved):

Issue #2334, opened 2025-06-03
bug: create app with only `okta.apps.manage` scope fails


exitcode0:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

<hr>

### Terraform Version
```
Terraform v1.11.4
on darwin_arm64
+ provider registry.terraform.io/okta/okta v4.18.0
```
### Affected Resource(s)
- `okta_app_*`
  - `okta_app_bookmark`
  - `okta_app_saml`
  - `okta_app_oauth`

### Terraform Configuration Files
```hcl
resource "okta_app_bookmark" "example_bookmark" {
  label = "example_bookmark - 123"
  url   = "https://example.com"
}
provider "okta" {
  org_name    = var.okta_org_name
  base_url    = var.okta_base_url
  client_id   = var.client_id
  scopes      = ["okta.apps.manage"]
  private_key = var.private_key
  # <Other attributes here as required>
}
```
### Error Output
```
Error:  
failed to set authentication policy for bookmark application:  
error finding default ACCESS_POLICY the API returned an error:  
The access token provided does not contain the required scopes.
```

### Expected Behavior

If the provider has the ability to `manage` okta applications, we should be able to create applications whilst also making use of the underlying default behavior of the Okta apps API, i.e. we should be able to create apps in the default authentication policy

### Actual Behavior

The provider creates the application as expected, but then explicitly updates the `authentication_policy` which then fails if we lack the permission to update policies, the relevant permission here is likely something like `okta.policies.manage`
I'm yet to validate this, but my expectation is that despite the error, the application is likely created regardless and would now be using the default authentication policy, if this is not the case the resource would likely be tainted

### Important Factoids
I think the problematic code is [here](https://github.com/okta/terraform-provider-okta/blob/master/okta/resource_okta_app_bookmark.go#L69-L74) I think we need to check if the user has provided a value for the attribute before we attempt to call `createOrUpdateAuthenticationPolicy()`

### References
N/A

aditya-okta:
@exitcode0 , but isn't the apps getting attached to the default Auth policy?

exitcode0:
[...]

PR #2363, opened 2025-06-22
disable call to auth policy api for service apps

aditya-okta:
Adds a new temporary field "disable_auth_policy" to prevent making calls to auth policy api for service apps. In future releases, we will be marking this field as deprecated.

### DIFF
--- a/okta/services/idaas/resource_okta_app_oauth.go
+++ b/okta/services/idaas/resource_okta_app_oauth.go
@@ -412,7 +412,7 @@ other arguments that changed will be applied.`,
 				Type:        schema.TypeBool,
 				Optional:    true,
 				Default:     false,
-				Description: "Disable adding authorization policy for SERVICE applications. WARNING: This is a temporary field since we will be removing the ability to toggle auth policy in the future. This is only for SERVICE applications, other application types will always have an auth policy attached to them.",
+				Description: "Disable adding authorization policy for SERVICE applications. WARNING: This is a temporary field. The ability to toggle the authorization policy will be removed in a future release. This field applies only to SERVICE applications, authorization policies will always be attached to other application types.",
 			},
 		}),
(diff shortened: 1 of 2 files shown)


okta/services/idaas/resource_okta_app_oauth.go
```diff
@@ -408,6 +408,12 @@ other arguments that changed will be applied.`,
 				Optional:    true,
 				Description: "URL reference to JWKS",
 			},
+			"disable_auth_policy": {
```
exitcode0, line 411: Do we need to expose this to the schema if we don't expect the user to ever need to configure this value?

exitcode0:
If we're going to do some work on this part of the code, we might also be able to address the root cause of #2334 at the same time

okta/services/idaas/resource_okta_app_oauth.go
aditya-okta: For new users, we require that no auth policy be attached to their service apps. To enforce this, they must explicitly set this field to true. For existing users, to avoid triggering unnecessary plan changes, they can omit this field—its default value is false.
Additionally, we've added a warning indicating that this field will be deprecated in the future, at which point we will completely stop attaching auth policies to service apps. Users can either opt in to this behaviour now by setting the field to true, or they will be required to update their configuration later when the field is removed.

exitcode0: I was just curious if we could avoid the schema change as it would then avoid having to deprecate the attribute again in the future 🤷

aditya-okta: i mean, the only motive behind adding this new field is to let -
1. The existing users know that they need to remove the auth policies from their Service apps.
2. Prevent the new users from attaching auth policy to their service apps.
Later, we will deprecate the field.

PR #2420, opened 2025-07-29
OKTA-817126 | Adding Missing Fields To Terraform State File To Avoid False Diffs For Token Type Inline Hooks

dhiwakar-okta:
OKTA-817126

### DIFF
--- a/okta/services/idaas/resource_okta_inline_hook_test.go
+++ b/okta/services/idaas/resource_okta_inline_hook_test.go
@@ -182,6 +182,64 @@ JSON
 	})
 }
 
+func TestAccResourceOktaInlineHook_token_channeljson(t *testing.T) {
+	resourceName := fmt.Sprintf("%s.token_channeljson", resources.OktaIDaaSInlineHook)
+	mgr := newFixtureManager("resources", resources.OktaIDaaSInlineHook, t.Name())
+	config := mgr.GetFixtures("okta_inline_hook_token_channeljson.tf", t)
+	acctest.OktaResourceTest(t, resource.TestCase{
+		PreCheck:                 acctest.AccPreCheck(t),
+		ErrorCheck:               testAccErrorChecks(t),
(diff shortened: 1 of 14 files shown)


okta/services/idaas/resource_okta_inline_hook_test.go
```diff
@@ -182,6 +182,64 @@ JSON
 	})
 }
 
+func TestAccResourceOktaInlineHook_token_channeljson(t *testing.T) {
```
monde, line 172: `TestAccResourceOktaInlineHook_token_channeljson` and `TestAccResourceOktaInlineHook_token_channel_auth` each only have one step in their acceptance test. This will only cover the `plan`, `apply (create)`, and `delete` actions. You'd need a second step `apply (update)` that changes the config.clientSecret, config.authScheme.value, auth.value values that this PR is targeted to address. Finally, you'd should have a third step which just corresponds to `plan` where you check the values are still the same, the run time will being doing another GET which will excercise the code changes here.

Take a look at this example for the three step acceptance test that focuses on verifying an update, pay attention to the checks on `password` and `recovery_answer`:

https://github.com/okta/terraform-provider-okta/blob/b437c780652f26062b175a8d8995776490f1635f/okta/services/idaas/resource_okta_user_test.go#L190-L240

monde (changes requested):

okta/services/idaas/resource_okta_inline_hook_test.go
dhiwakar-okta: Addressed, I've had to add 
func (s *secretHolder) CheckSecret(value string) error {
....
}
in order to verify the secret. The verification itself is a bit crude i.e. a single substring check but it is somewhat better than checking if the attribute merely exists. I couldn't think of a better way than to create a receiver type for the function CheckSecret. If you feel it's overkill, please do let me know and I can revert some of these changes & keep it simple.

monde (approved):
👍 :shipit:

PR #2443, opened 2025-08-21
OKTA-1002263 | OKTA-1002711 | Terraform Resource 'okta_app_signon_policy_rule' Errors Out Incorrectly When reauthenticateIn Is Set For chains Only

dhiwakar-okta:
ISSUE : 
When `reauthenticateIn` is set for one or more entries in the `chains` array, the resource creation fails. 

RCA : 
This is because in okta/services/idaas/resource_okta_app_signon_policy_rule.go we set this value for `AccessPolicyRule.Actions.AppSignOn.VerificationMethod.Type` by default.

FIX : 
Set the value conditionally.


aditya-okta (approved):
LGTM

Issue #2441, opened 2025-08-19
okta_idp_oidc fails to plan if excludeAdmins is checked in Dashboard


cthos:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
        Before submitting a bug report, we ask that you first [search existing issues](https://github.com/okta/terraform-provider-okta/issues) and [pull requests](https://github.com/okta/terraform-provider-okta/pulls) to see if someone else may have experienced the same issue or may have already submitted a fix for it. This helps to keep all relevant information in one place, including any potential workarounds.

### A Note on Terraform Core Issues

We also ask that you consider whether your issue may be related to Terraform Core. If you are running into one of the following scenarios, we recommend [opening an issue](https://github.com/hashicorp/terraform/issues/new/choose) in the Terraform Core repository instead:

* [Configuration Language](https://developer.hashicorp.com/terraform/language) or resource ordering issues
* [State](https://developer.hashicorp.com/terraform/language/state) and [State Backend](https://developer.hashicorp.com/terraform/language/backend) issues
* [Provisioner](https://developer.hashicorp.com/terraform/language/resources/provisioners/syntax) issues
* [Registry](https://registry.terraform.io/) issues
* Issues that span resources across multiple providers

### Terraform Version & Okta Provider Version(s)

Terraform v1.12.2
on linux_amd64
+ provider registry.terraform.io/okta/okta v5.3.0

### Affected Resource(s)

okta_idp_oidc (probably okta_idp_saml as well)

### Can this be done in the Admin UI?

Yes

### Can this be done in the actual API call?

Yes

### Customer Information

Organization Name: cthonicstudios
Paid Customer: no (though I am working on behalf of one and can confirm it happens on both kinds of orgs)

### Terraform Configuration

```tf
terraform {
  required_providers {
    okta = {
      source = "okta/okta"
      version = "~> 5.3.0"
    }
  }
}

provider "okta" {
  org_name = "integrator-7065415"
  base_url = "okta.com"
  client_id = "[REDACTED]"
  private_key_id = "[REDACTED]"
  private_key = "./okta-private-key.pem"
  scopes = [
[...]

PR #2457, opened 2025-08-26
[#2441][Bugfix] Fixes crash when additional account link settings are added

cthos:
…via the console.

As mentioned in #2441 there is an assumption that if `idp.Policy.AccountLink.Filter` is set then `idp.Policy.AccountLink.Filter.Groups.Include` is also set, but this is not true if you manually set other filter items via the console (like `excludeAdmins`).

Ideally the provider would also be able to set these params, and I'd potentially be willing to implement said feature, but for now I'd prefer that the plugin stop crashing on me.

### DIFF
--- a/okta/services/idaas/resource_okta_idp_oidc.go
+++ b/okta/services/idaas/resource_okta_idp_oidc.go
@@ -205,7 +205,7 @@ func resourceIdpRead(ctx context.Context, d *schema.ResourceData, meta interface
 	}
 	if idp.Policy.AccountLink != nil {
 		_ = d.Set("account_link_action", idp.Policy.AccountLink.Action)
-		if idp.Policy.AccountLink.Filter != nil {
+		if idp.Policy.AccountLink.Filter != nil && idp.Policy.AccountLink.Filter.Groups != nil {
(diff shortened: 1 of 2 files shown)


cthos:
I don't think I can add acceptance tests for this one as that would require updating the IdP via the API but I didn't see any contrivances for doing that in there.

dhiwakar-okta:
@cthos
Thanks for being patient with us.
The PR is missing VCR tests, please add those. You can refer [here](https://github.com/okta/terraform-provider-okta/blob/master/.github/CONTRIBUTING.md) for more information
The CLA also needs to be signed, more information about that can be found [here](https://developer.okta.com/cla/)
Please refer to https://github.com/okta/terraform-provider-okta?tab=readme-ov-file#contributing for a more comprehensive checklist.

Alternatively, I can take all of this up, including the code changes in a separate PR.

cthos:
I've emailed a signed CLA in, but I could use some guidance on how you'd like the test structured.

The reproduction case currently requires you to take a manual action in the Okta console, because the condition that triggers the bug is toggling on the setting to not allow an IDP to autojoin admin accounts.

I suppose I could have a test step that calls the API directly to make the setting in question? What do you think?

Issue #1855, opened 2023-12-18
Wrong definition of email_domain_id in brand resource


MichaelOtte-lhsystems:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version
1.6.6

### Affected Resource(s)

- okta_brand

### Terraform Configuration Files

```hcl
resource "okta_brand" "brand" {
  name                           = "Custom"
  custom_privacy_policy_url      = var.okta_brand.custom_privacy_policy_url 
  remove_powered_by_okta         = var.okta_brand.remove_powered_by_okta 
  email_domain_id                = var.email_domain_id
}
```

### Debug Output

```
  terrafrom validate 

Error: Invalid Configuration for Read-Only Attribute
│ 
│   with okta_brand.brand,
│   on brands.tf line 36, in resource "okta_brand" "brand":
│   36:   email_domain_id                = var. email_domain_id
│ 
│ Cannot set value for this attribute as the provider has marked it as read-only. Remove the configuration line setting the value.
│ 
│ Refer to the provider documentation or contact the provider developers for additional information about configurable and
│ read-only attributes that are supported.
```

### Expected Behavior

terraform apply will change the "email_domain_id" for brand so it should not be defined as Read-Only

### Steps to Reproduce

1. `terraform validate`

### References

- #0000

mvitt:
Hey @MichaelOtte-lhsystems,

This was changed from argument to attribute for the `okta_brand` resource because it you have the `email_domain_id` argument  set and have `brand_id ` argument set on the `okta_email_domain` resource you will get a Terraform cyclical error.

The way this should be done is:

1. Set the brand without `email_domain_id`.
2. Grab the brand id from the output of of step 1.
3. On the `okta_email_domain` set the `brand_id` with the id you retrieve from step 2.

MichaelOtte-lhsystems:
I recognised that there was a change https://github.com/okta/terraform-provider-okta/issues/1824 but how could the email_domain_id attached to brand if you have multiple brands as okta_email_domain only support one brand_id?


`
resource "okta_brand" "example1" {
  name = "example1
}
[...]

PR #2462, opened 2025-08-29
fix: don't update optional, unset fields of okta_brand

ohookins:
Potentially broken in https://github.com/okta/terraform-provider-okta/issues/1855 / https://github.com/okta/terraform-provider-okta/pull/2008 - when we update the `okta_brand` resource, it seems that the provider is a little too aggressive at setting the fields in the update brand request function. It includes fields that are optional, and potentially unset.

Unfortunately when setting the fields from the model, even if the field is unset coming back from the API representation, it ends up being set to the zero value instead and thus we over-eagerly update fields. The provider should only set fields we are telling it to, or preserve existing non-null values set automatically by the API.

We see this in the Terraform plan:

```
  # module.okta_core.module.okta_tenant_xxxx.okta_brand.this will be updated in-place
  ~ resource "okta_brand" "this" {
      ~ agree_to_custom_privacy_policy = false -> (known after apply)
      ~ brand_id                       = "xxxx" -> (known after apply)
      + custom_privacy_policy_url      = (known after apply)
      + default_app_app_instance_id    = "xxxx"
      + default_app_app_link_name      = "xxxx"
      ~ email_domain_id                = "xxxx" -> (known after apply)
        id                             = "xxxx"
      ~ is_default                     = false -> (known after apply)
      ~ links                          = jsonencode(
            {
              - emailDomain = {
                  - hints = {
                      - allow = [
                          - "GET",
                          - "PUT",
                        ]
                    }
                  - href  = "https://xxxx.oktapreview.com/api/v1/email-domains/xxxx"
                }
              - self        = {
                  - hints = {
                      - allow = [
                          - "GET",
                          - "PUT",
                          - "DELETE",
                        ]
                    }
                  - href  = "https://xxxx.oktapreview.com/api/v1/brands/xxxx"
                }
              - themes      = {
                  - hints = {
                      - allow = [
                          - "GET",
                        ]
                    }
                  - href  = "https://xxxx.oktapreview.com/api/v1/brands/xxxx/themes"
                }
            }
        ) -> (known after apply)
        name                           = "xxx"
        # (1 unchanged attribute hidden)
    }
```

### DIFF
--- a/okta/services/idaas/resource_okta_brand.go
+++ b/okta/services/idaas/resource_okta_brand.go
@@ -337,25 +337,51 @@ func buildCreateBrandRequest(model brandResourceModel) (okta.CreateBrandRequest,
 }
 
 func buildUpdateBrandRequest(model brandResourceModel, emailDomainID *string) (okta.BrandRequest, error) {
+	// mandatory fields
+	brandRequest := okta.BrandRequest{
+		Name:          model.Name.ValueStringPointer(),
+		EmailDomainId: emailDomainID,
+	}
+
+	// Only include optional fields if they are not null
+	if !model.AgreeToCustomPrivacyPolicy.IsNull() {
+		brandRequest.AgreeToCustomPrivacyPolicy = model.AgreeToCustomPrivacyPolicy.ValueBoolPointer()
+	}
+
+	if !model.CustomPrivacyPolicyURL.IsNull() {
+		brandRequest.CustomPrivacyPolicyUrl = model.CustomPrivacyPolicyURL.ValueStringPointer()
+	}
+
+	if !model.Locale.IsNull() {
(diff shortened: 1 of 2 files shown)


aditya-okta:
@ohookins , can you please fix the VCR test and sign the [Okta's CLA Agreement](https://developer.okta.com/cla/) as well if not signed.

ohookins:
Actually now that I've been playing with this a bit more, I'm not positive my assessment of the problem, and the solution, are correct. I suspect more fields are set automatically from the API, and the provider strips them out from the update which causes Terraform to think that a lot is changing.

Anyway I'll keep on working on this.

ohookins:
Not confident it solves the problem I'm experiencing, will close for the moment.

PR #2545, opened 2025-10-30
Update golangci yml

pranav-okta:
This PR updates the existing `.golangci.yml` with new linters.

Use `make fmt` to format the files using linters from `golangci-lint`.
Use `make tf-fmt` to format the example `.tf` files using `tfproviderint`.

### Notes:

The `make qc` uses the `tfproviderlint`, but it doesn't report any errors even though there are quite a few.


.golangci.yml
```diff
@@ -1,4 +1,52 @@
...
+    - gofmt
+    - goimports
+
+output:
+  formats:
+    text:
+      print-linter-name: true
+      print-issued-lines: true
+# issues:
+#   exclude:
```
dhiwakar-okta: Remove these comments if they aren't necessary.

GNUmakefile
```diff
@@ -89,26 +90,17 @@ smoke-test-play-vcr-acc:
...
-fmtcheck:
-	@gofumpt -d -l .
+.PHONY: fmt
+fmt: check-golangci-lint
+	@$(GOLANGCI_LINT) fmt
 
-errcheck:
-	@sh -c "'$(CURDIR)/scripts/errcheck.sh'"
+check-golangci-lint:
+	@which $(GOLANGCI_LINT) > /dev/null || curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b $(shell go env GOPATH)/bin $(GOLANGCI_LINT_VERSION)
```
dhiwakar-okta, line 103: This file https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh, I'm guessing we don't have control over where it's hosted, can we work with our internal team to maybe host it internally ? Please discuss with @manmohan-shaw-okta  once. My concern is that for any reason if the site becomes inaccessible, our pipeline would probably fail due to some kind of external dependency. I'll approve this PR if everything else looks good but please discuss this.

examples/data-sources/okta_campaign/data-source.tf
```diff
@@ -1,3 +1,3 @@
-data "okta_campaign" test{
+data "okta_campaign" "test" {
```
dhiwakar-okta, line 2: Do we really not require any argument to retrieve a campaign @aditya-okta  ? 
https://developer.okta.com/docs/api/iga/openapi/governance.api/tag/Campaigns/#tag/Campaigns/operation/getCampaign
PTAL this @aditya-okta ^

examples/data-sources/okta_entitlement/data-source.tf
```diff
@@ -1,4 +1,4 @@
-data "okta_entitlement" test{
+data "okta_entitlement" "test" {
```
dhiwakar-okta, line 2: Do we really not require any argument to retrieve an entitlement @aditya-okta ?
https://developer.okta.com/docs/api/iga/openapi/governance.api/tag/Entitlements/#tag/Entitlements/operation/getEntitlement
PTAL this @aditya-okta ^

dhiwakar-okta (dismissed):
Approved with comments

GNUmakefile
pranav-okta: [This](https://golangci-lint.run/docs/welcome/install/#other-ci) is the recommended way to add golangci-lint to your CI workflow, and a lot of other projects use a similar workflow. So, I don't think it will create any problems unless the maintainers lose their minds.

dhiwakar-okta (approved):

PR #2548, opened 2025-11-01
GH 2509 | OKTA-1040937 | okta_app_user_schema_property Reports Error During Creation Via Terraform

### LINKED ISSUE #2509: Root object was present, but now absent
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
        Before submitting a bug report, we ask that you first [search existing issues](https://github.com/okta/terraform-provider-okta/issues) and [pull requests](https://github.com/okta/terraform-provider-okta/pulls) to see if someone else may have experienced the same issue or may have already submitted a fix for it. This helps to keep all relevant information in one place, including any potential workarounds.

### A Note on Terraform Core Issues

We also ask that you consider whether your issue may be related to Terraform Core. If you are running into one of the following scenarios, we recommend [opening an issue](https://github.com/hashicorp/terraform/issues/new/choose) in the Terraform Core repository instead:

* [Configuration Language](https://developer.hashicorp.com/terraform/language) or resource ordering issues
* [State](https://developer.hashicorp.com/terraform/language/state) and [State Backend](https://developer.hashicorp.com/terraform/language/backend) issues
* [Provisioner](https://developer.hashicorp.com/terraform/language/resources/provisioners/syntax) issues
* [Registry](https://registry.terraform.io/) issues
* Issues that span resources across multiple providers

### Terraform Version & Okta Provider Version(s)

Terraform v1.9.2
on Cloud
+ provider registry.terraform.io/providers/okta/okta/ v6.2.0

### Affected Resource(s)

okta_app_user_schema_property

### Can this be done in the Admin UI?

Yes

### Can this be done in the actual API call?

Yes

### Customer Information

Organization Name: VaxCare
Paid Customer: Yes

### Terraform Configuration

```tf
resource "okta_app_user_schema_property" "prod_profile_active_directory" {
    for_each = {
        for i,j in local.profile_attributes_ad : i => j
        if try(j.base, false) == false
    }

    provider           = okta.prod
    app_id             = data.okta_app_oauth.prod_oauth["myapp"].id
    array_enum         = null
    array_type         = try(each.value.array_type, null)
    description        = try(each.value.description, null)
    enum               = null
    external_name      = each.value.title
    external_namespace = null
    index              = each.key
    master             = "PROFILE_MASTER"
    max_length         = try(each.value.max_length, null)
    min_length         = try(each.value.min_length, null)
    permissions        = try(each.value.permissions, "READ_ONLY")
    required           = try(each.value.required, false)
    scope              = try(each.value.scope, "NONE")
    title              = each.value.title
    type               = each.value.type
    union              = false
    unique             = null
    user_type          = "default"

}

### Debug Output

tbd

### Expected Behavior

Attribute created in Okta and in tf state without an error.

### Actual Behavior

The attribute does properly get created in Okta, so it works even though tf reports a failure. Re-running tries to re-add the field and fails with the same error. After doing a manual terraform import command from CLI, that changes the tf operation from create to modify, but still results in the same problem.

### Steps to reproduce

1. `terraform apply`

### Important Factoids

_No response_

### References

_No response_

### PR DESCRIPTION
**Pull request by dhiwakar-okta**
Fixes #2509 
There was a bug in the logic when suppressing 404.

### CODE REVIEW

**Review by aditya-okta — APPROVED** [review]

File: okta/services/idaas/resource_okta_app_user_custom_schema_property.go [hunk 1]
```diff
@@ -238,9 +238,9 @@ func updateAppUserSubSchemaProperty(ctx context.Context, d *schema.ResourceData,
 	retypeUserSchemaPropertyEnums(custom)
 	boc := utils.NewExponentialBackOffWithContext(ctx, 10*time.Second)
 	err = backoff.Retry(func() error {
-		_, resp, err := getOktaClientFromMetadata(meta).UserSchema.
-			UpdateApplicationUserProfile(ctx, d.Get("app_id").(string), *custom)
-		if utils.SuppressErrorOn404(resp, err) == nil {
+		_, resp, err := getOktaClientFromMetadata(meta).UserSchema.UpdateApplicationUserProfile(ctx, d.Get("app_id").(string), *custom)
+		// if error is nil, we skip this check entirely since utils.SuppressErrorOn404 will return nil either when err is nil or when resp.StatusCode is 404.
+		if err != nil && utils.SuppressErrorOn404(resp, err) == nil {
```
exitcode0, line 243:
```suggestion
		if err != nil || utils.SuppressErrorOn404(resp, err) == nil {
```
Should this be an `OR` rather than an `AND`

reasoning
```go
if err != nil && utils.SuppressErrorOn404(resp, err) == nil {
  // this code can never be reached
}

if err != nil || utils.SuppressErrorOn404(resp, err) == nil {
  // this code can be reached if either:
  // 1. err != nil
  // 2. utils.SuppressErrorOn404(resp, err) == nil
}
```

dhiwakar-okta: With the condition that you're proposing

```
if err != nil || utils.SuppressErrorOn404(resp, err) == nil {
		............
}

```

1. Let's say err != nil  evaluates to true
	since in your case it is logical OR, there is no point in proceeding further, we set an empty ID

2. Let's say err != nil evaluates to false (in other words, there is no error, it's a 2xx status code)
 	since in your case it is logical OR, we will try to evaluate the 2nd part of the condition i.e. `utils.SuppressErrorOn404(resp, err)`
	This is how we call the function --> utils.SuppressErrorOn404(resp, err) 
	Here's the rather small function pasted below for your reference. 

```
// Useful shortcut for suppressing errors from Okta's SDK when a resource does not exist. Usually used during deletion
// of nested resources.
func SuppressErrorOn404(resp *sdk.Response, err error) error {
    if resp != nil && resp.StatusCode == http.StatusNotFound {
        return nil
    }
    return ResponseErr(resp, err)
}
```

If you go through this function substituting nil in place of err, you'll see that eventually it evaluates to nil == nil which is true.
	What we're doing is setting an empty ID when err is in fact nil !

Changing it to && avoids this scenario, we set empty ID only in case of 404.
In case the error is genuine, we do have a block of code further down the line that should take care of it for attributes, finally we return err to what it was set.


If you still believe this is wrong, we can certainly discuss this further but the tests confirm this works.

PR #2571, opened 2025-11-20
feat: use mutex locks in `okta_app_signon_policy_rule`

pranav-okta:
This fix mirrors the solution in #1248, which addressed the same issue for okta_auth_server_policy_rule. That PR demonstrated that mutex locking at the resource level effectively prevents API concurrency issues while maintaining correct Terraform state management.

When managing multiple `okta_app_signon_policy_rule` resources with concurrent operations, the Okta API may encounter concurrency issues. While this provider implements internal locking to prevent conflicts within a single Terraform process, you should use explicit `depends_on` references between rules to ensure proper sequencing, especially when managing rule priorities.


okta/services/idaas/resource_okta_app_signon_policy_rule.go
```diff
@@ -401,6 +409,9 @@ func resourceAppSignOnPolicyRuleDelete(ctx context.Context, d *schema.ResourceDa
 		return resourceOIEOnlyFeatureError(resources.OktaIDaaSAppSignOnPolicyRule)
 	}
 
+	oktaMutexKV.Lock(resources.OktaIDaaSAppSignOnPolicyRule)
```
dhiwakar-okta, line 412: Please record VCR tests for the issue i.e record the test for the configuration that was buggy which your PR fixes.

docs/resources/app_signon_policy_rule.md
```diff
@@ -13,7 +13,11 @@ description: |-
 # Resource: okta_app_signon_policy_rule
 
 Manages a sign-on policy rules for the application.
+
 ~> **WARNING:** This feature is only available as a part of the Identity Engine. [Contact support](mailto:dev-inquiries@okta.com) for further information.
+
+~> **WARNING:** When managing multiple `okta_app_signon_policy_rule` resources with concurrent operations, the Okta API may encounter concurrency issues. While this provider implements internal locking to prevent conflicts within a single Terraform process, you should use explicit `depends_on` references between rules to ensure proper sequencing, especially when managing rule priorities.
```
dhiwakar-okta, line 19: Can we please include this caveat in the PR's description ? 
Just in case they don't see the Files Changed.

dhiwakar-okta (dismissed):
Approved with comments.

dhiwakar-okta (approved):
LGTM 🚀

PR #2595, opened 2025-12-16
OKTA-1001373 Add authenticator methods to `okta_authenticator`

pranav-okta:
This PR updates the SDK to v6 for `okta_authenticator` and adds methods to support `voice`, `sms`, `email`, etc.,


docs/data-sources/authenticator.md
```diff
@@ -35,6 +35,7 @@ data "okta_authenticator" "test_1" {
 
 ### Read-Only
 
+- `method` (Set of Object) Method-level configuration for authenticators that support it (phone_number, okta_verify, custom_otp). (see [below for nested schema](#nestedatt--method))
```
dhiwakar-okta, line 38: What is a Set of Object and how is it different from an Object ?

docs/data-sources/authenticator.md
```diff
@@ -45,4 +46,13 @@ data "okta_authenticator" "test_1" {
 - `status` (String) Status of the Authenticator.
 - `type` (String) Type of the authenticator
 
+<a id="nestedatt--method"></a>
+### Nested Schema for `method`
+
+Read-Only:
+
+- `type` (String) The authentication method type.
```
dhiwakar-okta: Are we talking about the authenticator type like app, email, federated etc. as described in the 200 response [here](https://developer.okta.com/docs/api/openapi/okta-management/management/tag/Authenticator/#tag/Authenticator/operation/getAuthenticatorMethod) ?

If yes, can we rewrite the description as _The type of authenticator method_ ?

docs/resources/authenticator.md
```diff
@@ -121,6 +182,31 @@ resource "okta_authenticator" "custom_app" {
...
+- `type` (String) The method type. Supported values depend on the authenticator key:
+  - For `phone_number`: `sms`, `voice`
+  - For `okta_verify`: `push`, `totp`, `signed_nonce`
+  - For `custom_otp`: `otp`
+
+Optional:
+
+- `status` (String) The method status: `ACTIVE` or `INACTIVE`. Default: `ACTIVE`
+- `settings` (String) Method-specific settings as a JSON string. Settings vary by method type:
+  - For `otp` method (custom_otp authenticator):
```
dhiwakar-okta, line 199: What about _factorProfileId_  & _verifiableProperties_ ?

docs/resources/authenticator.md
```diff
@@ -121,6 +182,31 @@ resource "okta_authenticator" "custom_app" {
...
+    - `protocol` (String) - Protocol type: `TOTP` or `HOTP`
+    - `encoding` (String) - Encoding format: `base32` or `base64`
+    - `algorithm` (String) - Hash algorithm: `HMacSHA1`, `HMacSHA256`, or `HMacSHA512`
+    - `timeIntervalInSeconds` (Number) - Time step in seconds (TOTP only)
+    - `passCodeLength` (Number) - Length of the OTP code
+    - `acceptableAdjacentIntervals` (Number) - Number of adjacent time intervals to accept
+  - For `signed_nonce` method (okta_verify authenticator):
+    - `algorithms` (Array) - Supported algorithms: `ES256`, `ES384`, `ES512`, `RS256`, `RS384`, `RS512`, `EdDSA`
+    - `keyProtection` (String) - Key protection level: `ANY`, `SOFTWARE`, or `HARDWARE`
```
dhiwakar-okta, line 208: What about _showSignInWithOV_ ?

docs/resources/authenticator.md
```diff
@@ -121,6 +182,31 @@ resource "okta_authenticator" "custom_app" {
 - `provider_type` (String) Provider type. Supported value for Duo: `DUO`. Supported value for Custom App: `PUSH`
 - `type` (String) he type of Authenticator. Values include: `password`, `security_question`, `phone`, `email`, `app`, `federated`, and `security_key`.
```
dhiwakar-okta: Please correct the typo here if you can **_he_** type... it must be **_The_** type :)

okta/services/idaas/data_source_okta_authenticator.go
```diff
@@ -88,83 +112,147 @@ func dataSourceAuthenticatorRead(ctx context.Context, d *schema.ResourceData, me
...
+				}
+
+				if config, ok := providerMap["configuration"].(map[string]interface{}); ok {
+					// Extract configuration based on authenticator type
+					switch authenticator.GetType() {
+					case "security_key":
+						if hostname, ok := config["hostName"].(string); ok {
+							_ = d.Set("provider_hostname", hostname)
+						}
+						if authPort, ok := config["authPort"].(float64); ok {
```
dhiwakar-okta, line 169: This should be fine since port numbers can't be decimals but just pointing out that  the type assertion to a float followed by a conversion to int here should be fine for the most part.

okta/services/idaas/resource_okta_authenticator.go
```diff
@@ -199,99 +235,181 @@ func resourceAuthenticatorCreate(ctx context.Context, d *schema.ResourceData, me
...
-	authenticator, _ := findAuthenticator(ctx, meta, d.Get("name").(string), d.Get("key").(string))
+	authenticator, _ := findAuthenticatorV6(ctx, client, d.Get("name").(string), d.Get("key").(string))
 	if authenticator == nil {
 		// otherwise hard create
-		authenticator, err = buildAuthenticator(d)
+		authenticatorReq, err := buildAuthenticatorV6(d, meta)
 		if err != nil {
 			return diag.FromErr(err)
 		}
 		activate := (d.Get("status").(string) == StatusActive)
```
dhiwakar-okta, line 248: This isn't part of your PR per se but "status" is optional, correct ?
If yes, it can be empty i.e nil and I think it might panic in case status isn't present, do we want to add a check to ensure status isn't nil or rather status is set ?

okta/services/idaas/resource_okta_authenticator.go
```diff
@@ -199,99 +235,181 @@ func resourceAuthenticatorCreate(ctx context.Context, d *schema.ResourceData, me
...
+				if err := json.Unmarshal(methodBytes, &methodUnion); err != nil {
+					return diag.FromErr(err)
+				}
+
+				// Update OTP method settings using ReplaceAuthenticatorMethod
+				methodReq := client.AuthenticatorAPI.ReplaceAuthenticatorMethod(ctx, authenticator.GetId(), "otp").
+					ListAuthenticatorMethods200ResponseInner(methodUnion)
+				_, methodResp, err := methodReq.Execute()
+				if err != nil {
+					logger(meta).Warn(fmt.Sprintf("Failed to set OTP settings: %v, request body: %s", err, string(methodBytes)))
```
dhiwakar-okta: Why is this just a warning and not something we fail on ?

okta/services/idaas/resource_okta_authenticator.go
```diff
@@ -199,99 +235,181 @@ func resourceAuthenticatorCreate(ctx context.Context, d *schema.ResourceData, me
...
-	if ok && authenticator.Status != status.(string) {
-		var err error
+	if ok && authenticator.GetStatus() != status.(string) {
 		if status.(string) == StatusInactive {
-			authenticator, _, err = getOktaClientFromMetadata(meta).Authenticator.DeactivateAuthenticator(ctx, d.Id())
+			_, resp, err := client.AuthenticatorAPI.DeactivateAuthenticator(ctx, d.Id()).Execute()
+			if err != nil {
+				return diag.Errorf("failed to change authenticator status: %v", err)
+			}
+			defer resp.Body.Close()
```
dhiwakar-okta: Are you actually using the resp anywhere, seems like you only check for the error, so why bother even deferring this ? you don't seem to use it so can't you just replace it with _ ?

okta/services/idaas/resource_okta_authenticator.go
```diff
@@ -199,99 +235,181 @@ func resourceAuthenticatorCreate(ctx context.Context, d *schema.ResourceData, me
...
 		} else {
-			authenticator, _, err = getOktaClientFromMetadata(meta).Authenticator.ActivateAuthenticator(ctx, d.Id())
+			_, resp, err := client.AuthenticatorAPI.ActivateAuthenticator(ctx, d.Id()).Execute()
+			if err != nil {
+				return diag.Errorf("failed to change authenticator status: %v", err)
+			}
+			defer resp.Body.Close()
```
dhiwakar-okta: Same comment as before

okta/services/idaas/resource_okta_authenticator.go
```diff
@@ -303,164 +421,835 @@ func resourceAuthenticatorDelete(ctx context.Context, d *schema.ResourceData, me
...
+// authenticatorMethod represents a method configuration
+type authenticatorMethod struct {
+	Type     string
+	Status   string
+	Settings map[string]interface{}
+}
+
+// supportsAuthenticatorMethods checks if the authenticator supports method-level configuration
+func supportsAuthenticatorMethods(key string) bool {
+	supportedKeys := map[string]bool{
```
dhiwakar-okta: Are you just looking to use a set ? If yes, prefer map[string]struct{} 
Empty structs don't consume any memory at all, it's slightly more efficient :) 
JFYI  --> https://www.bytesizego.com/blog/set-in-golang

okta/services/idaas/resource_okta_authenticator.go
```diff
@@ -303,164 +421,835 @@ func resourceAuthenticatorDelete(ctx context.Context, d *schema.ResourceData, me
...
+		} else {
+			// Check if status changed
+			if current.Status != desired.Status {
+				needsUpdate = true
+			}
+			// Check if settings changed (if settings are provided)
+			if desired.Settings != nil {
+				needsUpdate = true
+			}
+		}
```
dhiwakar-okta, line 968: Probably a single if condition like so would be easier to read right ? 

else if current.Status != desired.Status || desired.Settings != nil {
          needsUpdate = true
}

okta/services/idaas/resource_okta_authenticator.go
```diff
@@ -303,164 +421,835 @@ func resourceAuthenticatorDelete(ctx context.Context, d *schema.ResourceData, me
...
+				}
+			}
+
+			// Validate keyProtection if present
+			if keyProtection, ok := settings["keyProtection"]; ok {
+				kpStr, ok := keyProtection.(string)
+				if !ok {
+					return fmt.Errorf("'keyProtection' in method '%s' settings must be a string", methodType)
+				}
+				validKeyProtection := map[string]bool{"ANY": true, "SOFTWARE": true, "HARDWARE": true}
```
dhiwakar-okta: Prefer a map[string]struct{}

dhiwakar-okta (dismissed):
Approved with comments.

okta/services/idaas/resource_okta_authenticator.go
pranav-okta: The status is set to active by default.

okta/services/idaas/resource_okta_authenticator.go
pranav-okta: This is done.

okta/services/idaas/resource_okta_authenticator.go
pranav-okta: This is done.

dhiwakar-okta (approved):
LGTM 🚀

okta/services/idaas/resource_okta_authenticator.go
```diff
@@ -303,164 +419,835 @@ func resourceAuthenticatorDelete(ctx context.Context, d *schema.ResourceData, me
...
-			Type: d.Get("provider_type").(string),
-			Configuration: &sdk.AuthenticatorProviderConfiguration{
-				Host:           d.Get("provider_host").(string),
-				SecretKey:      d.Get("provider_secret_key").(string),
-				IntegrationKey: d.Get("provider_integration_key").(string),
-				UserNameTemplate: &sdk.AuthenticatorProviderConfigurationUserNamePlate{
-					Template: d.Get("provider_user_name_template").(string),
-				},
-			},
+		if !reflect.DeepEqual(oldValue, newValue) {
```
aditya-okta, line 465: can't we compare without reflect? since using reflect will be a bit slower

aditya-okta (approved):
approved with comments

okta/services/idaas/resource_okta_authenticator.go
pranav-okta: Yes, we can. The simpler approach is to use JSON with marshalling and unmarshalling. But it would require more changes and is a bit verbose. I noticed we are using `reflect`in a few other places

PR #2596, opened 2025-12-17
feat: Add Location header sanitization to VCR hook

exitcode0:
### Changes
- Adds `Location` header sanitization to the VCR hook, matching the existing pattern used for `Link` headers
- Prevents real org hostnames from leaking into cassette files via `Location` headers in `201 Created` responses

### Context
The `vcrHook` sanitizes hostnames in request URLs, request/response bodies, and `Link` headers, but does not sanitize `Location` headers

A `HTTP 201 Created` response includes a `Location` header which points to the newly created resource, this header contains the real org hostname (e.g. `acme-admin.oktapreview.com`)

### References
- #2740
- #2741

### DIFF
--- a/okta/acctest/acctest.go
+++ b/okta/acctest/acctest.go
@@ -547,6 +547,16 @@ func vcrHook(mgr *vcrManager) func(*cassette.Interaction) error {
 			i.Response.Headers.Add("Link", val)
 		}
 
+		// Sanitize Location header (used in 201 Created responses)
+		// %s/example-admin.okta.com/test-admin.dne-okta.com/
+		headerLocations := replaceHeaderValues(i.Response.Headers["Location"], orgAdminHostname, vcrAdminHostname)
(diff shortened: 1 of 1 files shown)


aditya-okta:
@exitcode0 Is there any particular use case for which this is required?

exitcode0:
> @exitcode0 Is there any particular use case for which this is required?

I was seeing circumstances where the  [\<domain>-admin.okta.com](\<domain>-admin.okta.com) url was being emitted to the VCR cassette

Maybe I was doing things wrong 🤷 
If others can't replicate, then we can close this PR

exitcode0:
it's possible that we're missing sanitization in other places as well, or that some people have run VCR recording incorrectly

here are some other examples - [link](https://github.com/search?q=repo%3Aokta%2Fterraform-provider-okta+-admin.okta+path%3A%2F%5Etest%5C%2Ffixtures%5C%2Fvcr%5C%2F%2F&type=code)

PR #2667, opened 2026-02-13
re-records vcr and adds changes to set settings when importing log_stream resource

### LINKED ISSUE #1929: `terraform import` for okta_log_stream imports settings as null
### Community Note

### Terraform Version

1.6.2

### Affected Resource(s)

- okta_log_stream

```hcl
resource "okta_log_stream" "eventbridge_okta_prod_useast1" {
  name   = "eventbridge_okta_prod_useast1"
  type   = "aws_eventbridge"
  status = "ACTIVE"
  settings {
    account_id        = "111111111111"
    region            = "us-east-1"
    event_source_name = "eventbridge_okta_prod_useast1"
  }
}

```

### Expected Behavior

The Log Stream resource should have imported with the settings (accountID, region, eventSourceName) populated

Yes

Yes

### Actual Behavior

The Log stream imported to state file with settings as a null value. After import, terraform plans state that the resource must be replaced.
Screenshot 2024-03-12 at 9 29 26 AM

### Steps to Reproduce

1. `terraform import okta_log_stream.eventbridge_okta_prod_useast1 {logStreamID}`

aditya-okta:
Fixes #1929 

### Root Cause
The applyLogStreamToState function had two issues:

Settings never written back to state: The function populated a local settings struct from the API response, but never assigned it back to `m.Settings`. The settings values were computed and then discarded.

`m.Settings.As()` fails silently during import: During terraform import, m.Settings is null (no prior state exists), so calling `m.Settings.As(ctx, settings, ...)` to deserialize it does nothing — the settings struct remains zero-valued. This also meant write-only sensitive values like the Splunk HEC token (which the API never returns) could not be preserved.

### Changes
resource_okta_log_stream.go

- Read prior settings safely: Before deserializing from m.Settings, check that it is not null/unknown (handles the import case where no prior state exists).

- Preserve write-only token: When the API response does not include the Splunk HEC token (it's write-only), preserve the value from the prior state to avoid "inconsistent values for sensitive attribute" errors.

- Write settings back to state: After populating the settings struct from the API response, construct a proper `types.Object` and assign it back to `m.Settings`. This is the core fix — without this, the computed settings were never persisted.

- Handle null attribute types during import: When m.Settings is null (import scenario), `m.Settings.AttributeTypes(ctx)` returns nil. Added a fallback that builds the attribute type map explicitly from the schema definition.

###VCR test fixtures
Re-recorded VCR cassettes for TestAccDataSourceOktaLogStream_read and TestAccResourceOktaLogStream_crud (classic and OIE) to match current API responses.

Before Fix: terraform import okta_log_stream.example <id> → settings imported as null → next plan forces replacement
After Fix: terraform import okta_log_stream.example <id> → settings populated correctly (account_id, region, event_source_name for EventBridge; edition, host for Splunk) → next plan shows no changes

### DIFF
--- a/okta/services/idaas/resource_okta_log_stream.go
+++ b/okta/services/idaas/resource_okta_log_stream.go
@@ -6,6 +6,7 @@ import (
 	"regexp"
 
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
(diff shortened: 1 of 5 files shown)


dhiwakar-okta (approved):
LGTM 🚀

PR #2680, opened 2026-02-22
adds a new resource for policy rules

aditya-okta:
### Summary
This PR introduces a new Terraform resource okta_app_signon_policy_rules that enables managing multiple app sign-on policy rules within a single resource block. This addresses long-standing issues with priority management and drift when using individual okta_app_signon_policy_rule resources.

### Problem Statement
When managing multiple app sign-on policy rules with individual okta_app_signon_policy_rule resources, users experience:

Priority drift: Okta's top-down priority synchronization causes unexpected priority shifts when rules are updated independently
Race conditions: Concurrent rule updates can lead to 409 conflicts and inconsistent state
Complex dependency management: Ordering rules correctly requires complex depends_on chains
Solution
The new okta_app_signon_policy_rules resource allows defining all rules for a policy in a single configuration block, ensuring:

1. Consistent priority ordering with atomic updates
2. Name-first matching for safe migration from individual rule resources
3. Automatic handling of rule reordering and name conflicts

### Key Features
1. Grouped rule management: Define all policy rules in a single resource
2. Name-first matching: Rules are identified by name (semantic key), preventing state mismatches
3. System rule support: Properly handles system rules (e.g., "Catch-all Rule") - they cannot be modified or deleted
4. Retry logic: Built-in retry for 429 (rate limit) and 503 (service unavailable) status codes
5. Migration support: Seamless migration path from okta_app_signon_policy_rule resources
6. Priority plan modifier: Uses UseStateForUnknown to prevent unnecessary diffs when priority is omitted


okta/services/idaas/resource_okta_app_signon_policy_rules.go
```diff
@@ -0,0 +1,1584 @@
...
+// sortRulesByPriority returns rules sorted by priority (ascending).
+// Rules without priority are placed at the end.
+func (r *appSignOnPolicyRulesResource) sortRulesByPriority(rules []policyRuleModel) []policyRuleModel {
+	sorted := make([]policyRuleModel, len(rules))
+	copy(sorted, rules)
+	sort.Slice(sorted, func(i, j int) bool {
+		iPriority := sorted[i].Priority
+		jPriority := sorted[j].Priority
+		if iPriority.IsNull() || iPriority.IsUnknown() {
+			return false
```
dhiwakar-okta, line 681: Why return false here when iPriority is null or unknown bit return true when jPriority is null or unknown ?

okta/services/idaas/resource_okta_app_signon_policy_rules.go
```diff
@@ -0,0 +1,1584 @@
...
+	var included, excluded []string
+
+	if !rule.UserTypesIncluded.IsNull() && !rule.UserTypesIncluded.IsUnknown() {
+		rule.UserTypesIncluded.ElementsAs(ctx, &included, false)
+	}
+	if !rule.UserTypesExcluded.IsNull() && !rule.UserTypesExcluded.IsUnknown() {
+		rule.UserTypesExcluded.ElementsAs(ctx, &excluded, false)
+	}
+
+	if len(included) > 0 || len(excluded) > 0 {
```
dhiwakar-okta, line 1201: condition checks either included > 0 or excluded > 0 but sets both of them, is that right ?

okta/services/idaas/resource_okta_app_signon_policy_rules.go
```diff
@@ -0,0 +1,1584 @@
...
+
+	if vm.InactivityPeriod != "" {
+		rule.InactivityPeriod = types.StringValue(vm.InactivityPeriod)
+	}
+
+	// Convert constraints to JSON strings.
+	if len(vm.Constraints) > 0 {
+		var constraintStrings []string
+		for _, constraint := range vm.Constraints {
+			if jsonBytes, err := json.Marshal(constraint); err == nil {
```
dhiwakar-okta, line 1420: If err != nil, it would simply append an empty slice of bytes, right ?

dhiwakar-okta (dismissed):
LGTM 🚀

okta/services/idaas/resource_okta_app_signon_policy_rules.go
aditya-okta: so basically, if the ith element has no priority it should come after j thus returns false(basically this pushes items with null/unknown priorities toward the end of the sorted list) , but if j has no priority and i has some priority then definitely, i should come before j, hence true

okta/services/idaas/resource_okta_app_signon_policy_rules.go
aditya-okta: yes, because the sdk has 
`type UserTypeCondition struct {
    Exclude []string `json:"exclude,omitempty"`
    Include []string `json:"include,omitempty"`
}`
the above structure, hence setting them both

dhiwakar-okta (dismissed):
LGTM 🚀

dhiwakar-okta (approved):
LGTM 🚀

PR #2684, opened 2026-02-24
fix: only set elCondition when custom_expression is not empty

pranav-okta:
## The Problem

  When creating an okta_app_signon_policy_rule resource without specifying custom_expression, the
  Terraform provider was sending an empty elCondition: {} object in the API request body. This
  caused Okta API validation errors.

## Root Cause

  In okta/services/idaas/resource_okta_app_signon_policy_rule.go, the buildAppSignOnPolicyRule
  function was always setting ElCondition regardless of whether custom_expression had a value:

```go
   // OLD CODE - Always set ElCondition
   ElCondition: &sdk.AccessPolicyRuleCustomCondition{
       Condition: d.Get("custom_expression").(string),
   },
```

Removed the `risk_score` check since risk_score is not enabled by default in the org. It is enabled when a feature flag is turned on.


okta/services/idaas/resource_okta_app_signon_policy_rule_test.go
```diff
@@ -51,7 +51,6 @@ func TestAccResourceOktaAppSignOnPolicyRule_crud(t *testing.T) {
 					resource.TestCheckResourceAttr(resourceName, "network_connection", "ANYWHERE"),
 					resource.TestCheckResourceAttr(resourceName, "constraints.#", "0"),
 					resource.TestCheckResourceAttr(resourceName, "re_authentication_frequency", "PT2H"),
-					resource.TestCheckResourceAttr(resourceName, "risk_score", "LOW"),
```
dhiwakar-okta, line 54: Why are we removing this check ?

dhiwakar-okta (approved):
Approved with comments.

okta/services/idaas/resource_okta_app_signon_policy_rule_test.go
pranav-okta: `risk_score` is not enabled by default in the org. It is enabled when a feature flag is turned on.

PR #2700, opened 2026-03-04
adds recovery if provider panics

aditya-okta:
### Summary
This PR adds comprehensive panic recovery to all Terraform provider resources and data sources. When a resource operation (Create, Read, Update, Delete) experiences an unhandled panic, the provider will now gracefully recover and return a diagnostic error instead of crashing the entire Terraform process.

### Problem
Provider panics (e.g., nil pointer dereferences, map assignment to nil) cause the entire Terraform process to crash mid-apply, leaving infrastructure in an inconsistent state. Users have to manually recover from these crashes, and the error messages provide limited context for debugging.

### Example of a panic that would previously crash Terraform:
`panic: assignment to entry in nil map
goroutine 118 [running]:
github.com/okta/terraform-provider-okta/okta/services/idaas.resourceAppOAuthCreate(...)`

### Solution
Wrap all resource and data source CRUD operations with defer recover() to catch panics and convert them into structured diagnostic errors with full stack traces.

After this change, the same panic produces:
`Error: Provider Crash in Create
  with okta_app_oauth.test_oauth,
  on main.tf line 22, in resource "okta_app_oauth" "test_oauth":
The provider crashed during the Create operation.
This is always a bug in the provider and should be reported to the provider developers.
Error: assignment to entry in nil map
Stack trace:
goroutine 118 [running]:
...`

### Key Benefits
1. No more Terraform crashes - Other resources continue to process even if one fails
2. Detailed error messages - Full stack traces included for debugging
3. Consistent behavior - Works for both Framework-based and SDK-based resources
4. Zero performance impact - Uses Go's native defer recover() pattern

### DIFF
--- a/examples/resources/okta_app_signon_policy_rule/basic.tf
+++ b/examples/resources/okta_app_signon_policy_rule/basic.tf
@@ -28,8 +28,8 @@ data "okta_app_signon_policy" "test" {
 }
 
 resource "okta_app_signon_policy_rule" "test" {
-  policy_id  = data.okta_app_signon_policy.test.id
-  name       = "testAcc_replace_with_uuid"
+  policy_id = data.okta_app_signon_policy.test.id
+  name      = "testAcc_replace_with_uuid"
   platform_include {
     os_expression = ""
     os_type       = "OTHER"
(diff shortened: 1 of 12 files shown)


dhiwakar-okta (approved):
LGTM 🚀

PR #2670, opened 2026-02-16
feat: adds okta_push_group data and resource

tim-koehler:
This PR adds `okta_push_group` resource as well as datasource. Additionally it also introduces the data source `okta_push_groups` which can be used to get all push groups associated with an application.

### Example Usage

```terraform
resource "okta_app_swa" "test" {
  app_links_json = jsonencode(
    {
      deelhr_link = true
    }
  )
  label                   = "Deel HR"
  preconfigured_app       = "deelhr"
  status                  = "ACTIVE"
  user_name_template_type = "BUILT_IN"
}
resource "okta_group" "test" {
  name = "Some Test Push Group"
}
resource "okta_push_group" "sample" {
  app_id                         = okta_app_swa.test.id
  source_group_id                = okta_group.test.id
  status                         = "ACTIVE"
  delete_target_group_on_destroy = true
}
```

Merging this closes #1409

PR #2712, opened 2026-03-09
Support AD as destination for okta_push_group

Dogers:
This follows on from #2670 - the `okta_push_group` resource doesn't correctly function with ActiveDirectory as a destination. This PR fixes that.

The initial commit also fixes an issue with the repo I hit initially - there are three folders under test/fixtures/vcr/governance/ that have a trailing space, causing git to fail on my Windows machines.

FYI @tim-koehler , as this builds on top of your recent PR I'm pinging you to see if you're able/willing to test it still works for your use case!


tim-koehler:
Hey @Dogers thanks for looking into that 🙏🏼 

So effectively you just added the `type` attribute that is marked optional in the API docs and then it worked? I didn't bother to set it because of that and was thinking that it doesn't make a difference.

Dogers:
The the type attribute was preventing creation, but the app_config attrib wasn't being saved to state, so caused continual changes too on a replan!

Dogers:
Rebased and now signed

dhiwakar-okta (approved):
LGTM 🚀

PR #2724, opened 2026-03-16
feat: add user risk API as a resource

pranav-okta:
This adds [User Risk](https://developer.okta.com/docs/api/openapi/okta-management/management/tags/userrisk) API to the Terraform for ITP support.


okta/services/idaas/resource_okta_user_risk.go
```diff
@@ -0,0 +1,136 @@
...
+			},
+		},
+	}
+}
+
+func resourceUserRiskCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	userId := d.Get("user_id").(string)
+	riskLevel := d.Get("risk_level").(string)
+
+	logger(meta).Info("setting user risk", "user_id", userId, "risk_level", riskLevel)
```
aditya-okta: is this required ?

okta/services/idaas/resource_okta_user_risk.go
```diff
@@ -0,0 +1,136 @@
+package idaas
```
aditya-okta, line 1: any reason we have gone ahead with plugin-sdk instead of the terraform-plugin-framework?

aditya-okta (changes requested):

okta/services/idaas/resource_okta_user_risk.go
pranav-okta: Migrated to terraform-plugin-framework

okta/services/idaas/resource_okta_user_risk.go
pranav-okta: Yes. If the risk_level is not one of the allowed values the API will fail and this logger will be helpful.

dhiwakar-okta (approved):
LGTM 🚀

aditya-okta (approved):

PR #2737, opened 2026-03-19
fix(request_condition): inconsistent result after apply

exitcode0:
- Fixes `"Provider produced inconsistent result after apply"` error on `okta_request_condition.priority` when the Okta API computes the `priority` value
  - `Create` & `Update`: stores the user's planned priority value in state
  - `Read` continues to use the API response so server-side changes surface as drift on the next `terraform plan` run

### DIFF
--- a/okta/services/governance/resource_request_condition.go
+++ b/okta/services/governance/resource_request_condition.go
@@ -2,6 +2,7 @@ package governance
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"time"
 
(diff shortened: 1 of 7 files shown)


okta/services/governance/resource_request_condition.go
```diff
@@ -218,11 +218,24 @@ func (r *requestConditionResource) Create(ctx context.Context, req resource.Crea
...
+	// server-side reassignment will surface as drift on the next plan via Read.
+	plannedPriority := data.Priority
+
 	resp.Diagnostics.Append(applyRequestConditionToState(ctx, &data, requestConditionResp)...)
 	if resp.Diagnostics.HasError() {
 		return
 	}
 
+	if !plannedPriority.IsNull() && !plannedPriority.IsUnknown() {
+		data.Priority = plannedPriority
```
aditya-okta, line 244: Can we add some kind of a warning when the values differ, we can use the below snippet
```
if !plannedPriority.IsNull() && !plannedPriority.IsUnknown() {
    if !data.Priority.Equal(plannedPriority) {
        resp.Diagnostics.AddWarning(
            "Priority reassigned by API",
            fmt.Sprintf("Requested priority %s was reassigned to %s by the server. "+
                "Run terraform plan to reconcile.", plannedPriority, data.Priority),
        )
    }
    data.Priority = plannedPriority
}
```

aditya-okta (changes requested):

exitcode0:
@aditya-okta I think this one should be ready for your re-review now

PR #2749, opened 2026-03-24
Entity risk policy and rules

pranav-okta:
This PR adds the `ENTITY_RISK` Policy rule to Terraform for ITP. For the Entity Risk Policy, we create a default policy. In this default policy, the system already generates a default catch-all rule with a priority of 99. Admins can add and edit any new rules within this default policy. They can't edit the default rule.

This also uses the latest SDK `v6.1.4`. While upgrading the latest SDK, there were a few failures related to Policies and Push Groups.


dhiwakar-okta (approved):
LGTM 🚀

Issue #2559, opened 2025-11-10
`okta_policy_rule_password` missing self service password requirments


zuzbel:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
* If an issue is assigned to a user, that user is claiming responsibility for the issue.
* [OKTA support]() Customers can ask to reach out to Okta Developer Support Engineer/Customer Success Engineer to expedite investigation and resolution of this issue.

### PROBLEM STATEMENT

Current resource `okta_policy_rule_password` is missing `selfServicePasswordReset`  (SSPR) requirments 
- Access control - determine if SSPR should be allowed after using specific authnetication policy  or Rule actions.
- establish the Rule actions of which authentication methods are required to initiate SSPR

These are supported via Okta API 
https://developer.okta.com/docs/api/openapi/okta-management/management/tag/Policy/#tag/Policy/operation/createPolicyRule but the resource itself does not account for these details

example of `POST /api/v1/policies/{policyId}/rules` body where `actions.selfServicePasswordReset.requirement` and `actions.selfServicePasswordReset.accessControl` are defined
```
{
    "name": "SSPR Rule",
    "priority": 1,
    "status": "ACTIVE",
    "conditions": {
      "people": {
        "users": {
          "exclude": []
        }
      },
      "network": {
        "connection": "ANYWHERE"
      }
    },
    "actions": {
      "passwordChange": {
        "access": "ALLOW"
      },
     "selfServicePasswordReset": {
            "access": "ALLOW",
            "requirement": {
                "primary": {
                    "methods": [
                        "otp"
                    ],
                    "methodConstraints": [
                        {
                            "method": "otp",
                            "allowedAuthenticators": [
                                {
                                    "key": "google_otp"
                                }
                            ]
                        }
                    ]
                },
                "stepUp": {
                    "required": true
                },
                "accessControl": "LEGACY"
            }
        },
      "selfServiceUnlock": {
        "access": "ALLOW"
      }
    },
    "system": false,
    "type": "PASSWORD"
  }
```

### PROPOSED ENHANCMENT
[...]

PR #2752, opened 2026-03-26
GH-2559 | Adding Self Service Password Requirements Support for okta_policy_rule_password 

dhiwakar-okta:
Add `step_up_methods` and SSPR requirement improvements to `okta_policy_rule_password`

Extends the password_reset_requirement block in `okta_policy_rule_password` to support step_up_methods — the list of authenticator methods (e.g. `security_question`) required for the secondary SSPR step when `step_up_enabled = true`. The Read function now correctly syncs all SSPR requirement fields (`primary_methods`, `method_constraints`, `step_up_enabled`, `step_up_methods`) back to state, preventing perpetual diffs. The `okta_policy_rule_password` data source has been updated to match, receiving the same schema additions and a simplified read path that looks up rules directly by `id` and `policy_id`. Acceptance tests and documentation have been updated to cover the new fields including a dedicated 4-step SSPR test and a new data source test.


okta/services/idaas/data_source_okta_policy_rule_password.go
```diff
@@ -0,0 +1,233 @@
...
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"name"},
+				Description:   "ID of the rule. Conflicts with `name`.",
+			},
+			"name": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"id"},
```
pranav-okta: Why `name` conflicts with the `id`?

okta/services/idaas/data_source_okta_policy_rule_password.go
```diff
@@ -0,0 +1,233 @@
...
+				Description: "Network zones to include (when `network_connection` = `ZONE`).",
+				Elem:        &schema.Schema{Type: schema.TypeString},
+			},
+			"network_excludes": {
+				Type:        schema.TypeList,
+				Computed:    true,
+				Description: "Network zones to exclude (when `network_connection` = `ZONE`).",
+				Elem:        &schema.Schema{Type: schema.TypeString},
+			},
+			"users_excluded": {
```
pranav-okta, line 60: Should we include `users_included`?

okta/services/idaas/data_source_okta_policy_rule_password.go
dhiwakar-okta: It shouldn't, I've removed the `ConflictsWith` property.

okta/services/idaas/data_source_okta_policy_rule_password.go
dhiwakar-okta: Thanks for bringing this up, I've added `users_included` as well as `groups_included`, `groups_excluded`

okta/services/idaas/data_source_okta_policy_rule_password.go
```diff
@@ -0,0 +1,254 @@
+package idaas
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
```
pranav-okta, line 7: I apologise for not bringing this up earlier. But is there a reason why we aren't using `terraform-plugin-framework` here?

okta/services/idaas/data_source_okta_policy_rule_password_test.go
```diff
@@ -0,0 +1,49 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet(dataSourceName, "id"),
+					resource.TestCheckResourceAttr(dataSourceName, "name", acctest.BuildResourceName(mgr.Seed)),
+					resource.TestCheckResourceAttr(dataSourceName, "status", "ACTIVE"),
+					resource.TestCheckResourceAttr(dataSourceName, "password_change", "ALLOW"),
+					resource.TestCheckResourceAttr(dataSourceName, "password_reset", "ALLOW"),
+					resource.TestCheckResourceAttr(dataSourceName, "password_unlock", "ALLOW"),
+					resource.TestCheckResourceAttr(dataSourceName, "password_reset_access_control", "LEGACY"),
+					// People conditions
+					resource.TestCheckTypeSetElemAttr(dataSourceName, "users_included.*", "00ustguf78owmG7Rt1d7"),
```
pranav-okta: This might fail if we run the tests under different tenants. Don't use hardcoded IDs.

okta/services/idaas/resource_okta_policy_rule_password.go
```diff
@@ -18,6 +20,24 @@ func resourcePolicyPasswordRule() *schema.Resource {
 		Importer:      createPolicyRuleImporter(),
 		Description:   "Creates a Password Policy Rule. This resource allows you to create and configure a Password Policy Rule.",
 		Schema: buildRuleSchema(map[string]*schema.Schema{
+			"users_included": {
```
pranav-okta, line 22: We have to add `users_excluded` to this resource.

okta/services/idaas/resource_okta_policy_rule_password.go
```diff
@@ -36,79 +56,445 @@ func resourcePolicyPasswordRule() *schema.Resource {
...
+
+	conds := v6okta.PasswordPolicyRuleConditions{}
+	conds.SetNetwork(networkCond)
+
+	// Build the people conditions (users and groups).
+	peopleCond := v6okta.PolicyPeopleCondition{}
+	hasPeople := false
+
+	userCond := v6okta.UserCondition{}
+	if exclude, ok := d.GetOk("users_excluded"); ok {
```
pranav-okta, line 353: We have to add `users_excluded` to the resource.

okta/services/idaas/resource_okta_policy_rule_password_test.go
```diff
@@ -193,10 +204,99 @@ resource "%s" "%s" {
 	name     = "%s"
 	priority = 999
 	status   = "ACTIVE"
+	users_included  = ["00ustguf78owmG7Rt1d7"]
```
pranav-okta, line 202: Don't use hardcoded IDs. Use data sources of user and group and use the IDs.

okta/services/idaas/resource_okta_policy_rule_password_test.go
```diff
@@ -115,6 +121,11 @@ func ensureRuleExists(resourceName string) resource.TestCheckFunc {
 }
 
 func checkRuleDestroy(ruleType string) func(*terraform.State) error {
+	if os.Getenv("OKTA_VCR_TF_ACC") == "play" {
```
pranav-okta: We aren't actually testing the deletion of the policy rule here. Can we remove this?

okta/services/idaas/resource_okta_policy_rule_password_test.go
```diff
@@ -95,6 +96,11 @@ func TestAccResourceOktaPolicyRulePassword_priority(t *testing.T) {
 }
 
 func ensureRuleExists(resourceName string) resource.TestCheckFunc {
+	if os.Getenv("OKTA_VCR_TF_ACC") == "play" {
```
pranav-okta, line 99: Do you think we can remove this?

okta/services/idaas/resource_okta_policy_rule_password.go
```diff
@@ -36,79 +56,445 @@ func resourcePolicyPasswordRule() *schema.Resource {
...
+	}
+	policyID := d.Get("policy_id").(string)
+	if policyID == "" {
+		return diag.Errorf("'policy_id' field should be set")
+	}
+
+	resp, err := getOktaV6ClientFromMetadata(meta).PolicyAPI.DeletePolicyRule(ctx, policyID, d.Id()).Execute()
 	if err != nil {
+		// Already gone — not an error.
+		if resp != nil && resp.Response != nil && resp.StatusCode == http.StatusNotFound {
```
pranav-okta: If the rule is not found, it can be an error. Why `404` is not an error here?

okta/services/idaas/resource_okta_policy_rule_password.go
```diff
@@ -36,79 +56,445 @@ func resourcePolicyPasswordRule() *schema.Resource {
...
+		return diag.FromErr(err)
+	}
+	policyID := d.Get("policy_id").(string)
+	if policyID == "" {
+		return diag.Errorf("'policy_id' field should be set")
+	}
+
+	rule := buildPolicyRulePassword(d)
+	// Wrap the typed rule in the union type required by the V6 PolicyAPI.
+	wrapper := v6okta.PasswordPolicyRuleAsListPolicyRules200ResponseInner(&rule)
```
pranav-okta: This is just a policy rule, right?  Can we rename this to `policyRule`?

okta/services/idaas/resource_okta_policy_rule_password.go
```diff
@@ -36,79 +56,445 @@ func resourcePolicyPasswordRule() *schema.Resource {
...
+
+	// Use exponential backoff to handle transient 500 errors from the Okta API.
```
pranav-okta: Do we need to retry for 500 errors?

okta/services/idaas/resource_okta_policy_rule_password.go
```diff
@@ -36,79 +56,445 @@ func resourcePolicyPasswordRule() *schema.Resource {
...
+	if err := ensureNotDefaultRule(d); err != nil {
+		return diag.FromErr(err)
+	}
+	policyID := d.Get("policy_id").(string)
+	if policyID == "" {
+		return diag.Errorf("'policy_id' field should be set")
+	}
+
+	rule := buildPolicyRulePassword(d)
+	wrapper := v6okta.PasswordPolicyRuleAsListPolicyRules200ResponseInner(&rule)
```
pranav-okta:
```suggestion
	policyRule := v6okta.PasswordPolicyRuleAsListPolicyRules200ResponseInner(&rule)
```

pranav-okta (changes requested):

okta/services/idaas/data_source_okta_policy_rule_password.go
dhiwakar-okta: We can consider migrating to the new framework later on for this particular data source.

okta/services/idaas/resource_okta_policy_rule_password.go
dhiwakar-okta: Done.

okta/services/idaas/resource_okta_policy_rule_password_test.go
dhiwakar-okta: Done.

okta/services/idaas/resource_okta_policy_rule_password.go
dhiwakar-okta: Done.

okta/services/idaas/resource_okta_policy_rule_password.go
dhiwakar-okta: Done.

okta/services/idaas/resource_okta_policy_rule_password.go
dhiwakar-okta: You're right this might not be needed, removing it.

okta/services/idaas/resource_okta_policy_rule_password.go
dhiwakar-okta: Will simply return if err when err != nil is True

okta/services/idaas/resource_okta_policy_rule_password.go
dhiwakar-okta: Done.

okta/services/idaas/data_source_okta_policy_rule_password.go
```diff
@@ -0,0 +1,254 @@
...
+		Description: "Get a Password Policy Rule from Okta.",
+		Schema: map[string]*schema.Schema{
+			"policy_id": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "ID of the Policy owning this rule.",
+			},
+			"id": {
+				Type:        schema.TypeString,
+				Optional:    true,
```
aditya-okta: Shouldn't this be required, since we need it while making the API call?

okta/services/idaas/resource_okta_policy_rule_password.go
```diff
@@ -36,79 +61,425 @@ func resourcePolicyPasswordRule() *schema.Resource {
...
+						methodConstraints = append(methodConstraints, map[string]interface{}{
+							"method":                 mc.GetMethod(),
+							"allowed_authenticators": allowedAuths,
+						})
+					}
+					reqMap["method_constraints"] = methodConstraints
+				}
+				if stepUp := req.StepUp; stepUp != nil {
+					reqMap["step_up_enabled"] = stepUp.GetRequired()
+					reqMap["step_up_methods"] = stepUp.Methods
```
aditya-okta, line 252: can we use GetMethods(), if its available?

okta/services/idaas/data_source_okta_policy_rule_password.go
```diff
@@ -0,0 +1,254 @@
...
+			_ = d.Set("password_change", pc.GetAccess())
+		}
+		if su := actions.SelfServiceUnlock; su != nil {
+			_ = d.Set("password_unlock", su.GetAccess())
+		}
+		if sspr := actions.SelfServicePasswordReset; sspr != nil {
+			_ = d.Set("password_reset", sspr.GetAccess())
+			if req := sspr.Requirement; req != nil {
+				if ac := req.AccessControl; ac != nil {
+					_ = d.Set("password_reset_access_control", ac)
```
aditya-okta, line 210: If the API stops returning accessControl (e.g., after switching away from the field), the old value stays in state permanently, causing a perpetual diff. Add an explicit clear when nil:


```
if ac := req.AccessControl; ac != nil {
    _ = d.Set("password_reset_access_control", *ac) // note: ac is *string
} else {
    _ = d.Set("password_reset_access_control", "")
}
```

aditya-okta (changes requested):

okta/services/idaas/data_source_okta_policy_rule_password.go
dhiwakar-okta: true, I'll fix it.

okta/services/idaas/resource_okta_policy_rule_password.go
dhiwakar-okta: Done.

aditya-okta (approved):

aditya-okta (dismissed):

okta/services/idaas/resource_okta_policy_rule_password.go
```diff
@@ -36,79 +62,428 @@ func resourcePolicyPasswordRule() *schema.Resource {
...
+
+	// Sync base rule fields.
+	_ = d.Set("name", rule.GetName())
+	_ = d.Set("status", rule.GetStatus())
+	_ = d.Set("priority", int(rule.GetPriority()))
+
+	// Sync network conditions.
+	if conds := rule.Conditions; conds != nil {
+		if net := conds.Network; net != nil {
+			_ = d.Set("network_connection", net.GetConnection())
```
pranav-okta, line 215: Network Connection values aren't defined in the schema. Can you confirm?

okta/services/idaas/resource_okta_policy_rule_password_test.go
pranav-okta: Is this still relevant?

pranav-okta (commented):
A few minor things before approval.

aditya-okta (approved):

pranav-okta (approved):

PR #2753, opened 2026-03-26
addresses the issue caused when using dynamic "rule" block

aditya-okta:
This PR addresses issue that was causing failure when attempting to use dynamic rule blocks whose contents come from computed locals, the provider throws a "Value Conversion Error" during the validate/plan phase.


okta/services/idaas/resource_okta_app_signon_policy_rules.go
```diff
@@ -327,29 +318,28 @@ func (r *appSignOnPolicyRulesResource) Create(ctx context.Context, req resource.
 		}
 		createdRules = append(createdRules, resultRule)
 	}
-
 	// Reorder to match config order (Terraform expects state order to match plan order).
-	plan.Rules = r.reorderRulesToMatchPlan(createdRules, plan.Rules)
+	reorderedRules := r.reorderRulesToMatchPlan(createdRules, rules)
 	plan.ID = plan.PolicyID
+	// Marshal back to types.List
+	plan.Rules, resp.Diagnostics = types.ListValueFrom(ctx, r.policyRuleObjectType(), reorderedRules)
```
pranav-okta: Don't assign this directly to `resp.Diagnostics`. We might lose the diagnostics we added earlier.

okta/services/idaas/resource_okta_app_signon_policy_rules.go
pranav-okta, line 101: Do we hit the same error if they use `dynamic_rule` for `PlatformInclude`?

okta/services/idaas/resource_okta_app_signon_policy_rules.go
```diff
@@ -196,29 +195,26 @@ func (t *nameTracker) updateMapping(ruleID, newName string) {
...
 	resp.Diagnostics.Append(req.Config.Get(ctx, &data)...)
 	if resp.Diagnostics.HasError() {
 		return
 	}
-
-	seen := make(map[string]int, len(data.Rules))
-	seenPriority := make(map[int64]int, len(data.Rules))
-	for i, rule := range data.Rules {
+	var rules []policyRuleModel
+	data.Rules.ElementsAs(ctx, &rules, false)
```
pranav-okta: This returns `diagnostics`, and we're not capturing or handling it. Is this intentional?

aditya-okta: added diagnostics here

okta/services/idaas/resource_okta_app_signon_policy_rules.go
aditya-okta: tested with a config like below
`      dynamic "platform_include" {
        for_each = rule.value.platform_include
        content {
          type          = platform_include.value.type
          os_type       = platform_include.value.os_type
          os_expression = try(platform_include.value.os_expression, null)
        }
      }`

Working fine.

okta/services/idaas/resource_okta_app_signon_policy_rules.go
aditya-okta: appending to diagnostics now instead of assigning directly.

pranav-okta (approved):

okta/services/idaas/resource_okta_app_signon_policy_rules.go
```diff
@@ -67,9 +66,9 @@ type appSignOnPolicyRulesResource struct {
 
 // appSignOnPolicyRulesModel represents the Terraform state for this resource.
 type appSignOnPolicyRulesModel struct {
-	ID       types.String      `tfsdk:"id"`
-	PolicyID types.String      `tfsdk:"policy_id"`
-	Rules    []policyRuleModel `tfsdk:"rule"`
+	ID       types.String `tfsdk:"id"`
+	PolicyID types.String `tfsdk:"policy_id"`
+	Rules    types.List   `tfsdk:"rule"`
```
dhiwakar-okta, line 71: This sort of a change would lead to a diff post upgrade right ?

dhiwakar-okta (approved):
LGTM 🚀

okta/services/idaas/resource_okta_app_signon_policy_rules.go
aditya-okta: I checked  by creating a new config using the existing code on master and then doing a plan on the feature branch, it doesn't show a diff.

PR #2759, opened 2026-03-29
feat: add session-violation-policy and rules as resource

pranav-okta:
Adds Terraform support for the `SESSION_VIOLATION_DETECTION` policy type, which is a system policy automatically provisioned when the Session Violation Detection feature is enabled on an Okta org.

##  New Resources

- `okta_session_violation_policy` (Data Source)
  Retrieves the system-created Session Violation Detection Policy, including the ID of its modifiable rule.

- `okta_session_violation_policy_rule` (Resource)
  Manages the modifiable rule of the Session Violation Detection Policy. Since the rule is pre-provisioned by Okta, this resource supports import and update only — it cannot be created or deleted.


docs/data-sources/session_violation_policy.md
```diff
@@ -0,0 +1,23 @@
...
+```terraform
+data "okta_session_violation_policy" "example" {
+}
+```
+
+## Attributes Reference
+
+- `id` - ID of the Session Violation Detection Policy.
+- `name` - Name of the Session Violation Detection Policy.
+- `status` - Status of the policy: `ACTIVE` or `INACTIVE`.
```
dhiwakar-okta, line 22: What about attributes for response fields such as `priority`, `system`, `created`, `conditions`

dhiwakar-okta (dismissed):
Approved with comments.

docs/data-sources/session_violation_policy.md
pranav-okta: `conditions` are already supported. We cannot update the `system` and `created` fields for this resource. 

However, `priority` can be updated, and I'll add this.

dhiwakar-okta (approved):
LGTM 🚀

PR #2771, opened 2026-04-08
allow unknown values in rule dynamic blocks

mbampi:
## Problem

ValidateConfig calls data.Rules.ElementsAs(ctx, &rules, false) with allowUnknown=false. When a dynamic rule block references any Set-typed attribute (groups_included, groups_excluded, users_included, users_excluded, device_assurances_included, user_types_included, user_types_excluded) using a managed resource ID (e.g. okta_group.*.id), that resource ID is unknown during config validation. Unlike types.List, a types.Set with unknown elements becomes entirely unknown (unknown values cannot be hashed into a Set). This causes ElementsAs(..., false) to fail with:

```
╷
│  Error: Value Conversion Error
│ 
│   with okta_app_signon_policy_rules.rules_v2,
│   on authn_policies_engine_v2.tf line 33, in resource "okta_app_signon_policy_rules" "rules_v2":
│   33: resource "okta_app_signon_policy_rules" "rules_v2" {
│ 
│ An unexpected error was encountered trying to build a value. This is always an error in the provider. Please report the following to the provider developer:
│ 
│ Received unknown value, however the target type cannot handle unknown values. Use the corresponding `types` package type or a custom type that handles unknown
│ values.
│ 
│ Path: 
│ Target Type: []idaas.policyRuleModel
│ Suggested Type: basetypes.ListValue
╵
```

##  Fix

Change allowUnknown from false to true in the single ElementsAs call inside ValidateConfig. The validation loop already handles unknowns correctly: it skips rules with unknown namess and guards the priority check (!rule.Priority.IsUnknown()). All other ElementsAs calls in the file remain false because they operate on state or apply-time values that are always concrete.

## Validation

It was validated locally by changing the provider locally, and using the local build with `dev_overrides`. From that I ran `terraform plan` and it ran as expected, with no errors.
Please feel free to validate internally however is required.

## Related PRs

https://github.com/okta/terraform-provider-okta/pull/2753

### DIFF
--- a/okta/services/idaas/resource_okta_app_signon_policy_rules.go
+++ b/okta/services/idaas/resource_okta_app_signon_policy_rules.go
@@ -252,7 +252,7 @@ func (r *appSignOnPolicyRulesResource) ValidateConfig(ctx context.Context, req r
 		return
 	}
 	var rules []policyRuleModel
-	resp.Diagnostics.Append(data.Rules.ElementsAs(ctx, &rules, false)...)
+	resp.Diagnostics.Append(data.Rules.ElementsAs(ctx, &rules, true)...)
 	if resp.Diagnostics.HasError() {
 		return
 	}


aditya-okta (approved):

aditya-okta:
@mbampi , can you sign the [Okta Individual Contributor License Agreement](https://developer.okta.com/cla/) and email to CLA@okta.com?

PR #2787, opened 2026-04-19
feat(identity_source): add resources, data sources, tests, and docs

aditya-okta:
Implements full Terraform support for the Okta Identity Source API:

- 6 resources: session, user, bulk_upsert, bulk_delete, bulk_groups_upsert, bulk_groups_delete
- 4 data sources: sessions, users, groups, group_memberships
- Acceptance tests with VCR cassettes for all 10 resources/data sources
- Example .tf fixtures and docs/resources + docs/data-sources markdown


okta/services/idaas/resource_okta_identity_source_group.go
```diff
@@ -0,0 +1,259 @@
...
+						Description: "Name of the group",
+						Optional:    true,
+					},
+				},
+			},
+		},
+	}
+}
+
+func (r *identitySourceGroupResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
```
pranav-okta, line 111: In all **_bulk.go** Read is no-op, Delete does nothing, ImportState is of no use. Can we have clear documentation what these resources are capable of?

okta/services/idaas/resource_okta_identity_source_group_membership.go
```diff
@@ -0,0 +1,239 @@
+// Code generated by tf-generator. DO NOT EDIT.
```
pranav-okta, line 1: I don't see a test file for this.

okta/services/idaas/resource_okta_identity_source_bulk_group_memberships_delete.go
```diff
@@ -0,0 +1,204 @@
+// Code generated by tf-generator. DO NOT EDIT.
```
pranav-okta, line 1: I don't see a test file for this.

okta/services/idaas/resource_okta_identity_source_user.go
```diff
@@ -0,0 +1,335 @@
...
+		body.SetProfile(*nestedProfile)
+	}
+	createReq = createReq.UserRequestSchema(*body)
+	_, err := createReq.Execute()
+	if err != nil {
+		resp.Diagnostics.AddError("Error creating identity_source_user", err.Error())
+		return
+	}
+	// Create returns no body (204); set ID from the plan field.
+	plan.ID = plan.ExternalId
```
pranav-okta, line 238: What is the difference between `ID` and `ExternalId`?

okta/services/idaas/resource_okta_identity_source_user.go
```diff
@@ -0,0 +1,335 @@
...
+			return
+		}
+		resp.Diagnostics.AddError("Error reading identity_source_user", err.Error())
+		return
+	}
+	// Map API response fields to state (scalar types only; WriteOnly fields are skipped — response type doesn't have them)
+	state.Created = types.StringValue(result.GetCreated().Format(time.RFC3339))
+	state.ExternalId = types.StringValue(result.GetExternalId())
+	state.LastUpdated = types.StringValue(result.GetLastUpdated().Format(time.RFC3339))
+	state.ID = types.StringValue(result.GetId())
```
pranav-okta, line 187: Are these two IDs the same?

pranav-okta (changes requested):
I have not reviewed the docs and tests yet. PTAL

okta/services/idaas/data_source_okta_identity_source_group_memberships.go
```diff
@@ -0,0 +1,215 @@
...
+			"Expected *config.Config, got something else. Please report this issue to the provider developers.",
+		)
+		return
+	}
+	d.Config = cfg
+}
+
+func (d *identitySourceGroupMembershipsDataSource) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		MarkdownDescription: "Retrieves the group memberships for the given identity source group in the given identity source instance (`GET /api/v1/identity-sources/{identitySourceId}/groups/{groupOrExternalId}/memberships`).",
```
dhiwakar-okta, line 62: Any reason why we're mentioning the underlying API in the description ? 

`(`GET /api/v1/identity-sources/{identitySourceId}/groups/{groupOrExternalId}/memberships`)`


Do other providers such as AWS, GCP, Datadog etc do this ?

okta/services/idaas/data_source_okta_identity_source_group_memberships.go
```diff
@@ -0,0 +1,215 @@
...
+					if c == '&' || c == '>' || c == ';' || c == ' ' {
+						return after[:i]
+					}
+				}
+				return after
+			}
+		}
+	}
+	return ""
+}
```
dhiwakar-okta, line 178: Can you please take a closer look at this implementation ? 
This string manipulation to extract the `after` URL, I'm sure there is some built in package that would do this for us.
Check out the implementation of `listResourceSetResources` in `resource_okta_resouce_set.go` and also explore packages in Go's standard library that can extract Headers, URLs etc. from standard HTTP Response objects. 

```
// APIResponse stores the API response returned by the server.
type APIResponse struct {
	*http.Response
	cli *APIClient
	pg  Pagination
}
```

Okta's `APIResponse` struct is a wrapper on top of `http.Response`.
If none of this works then let's retain this logic BUT let's try to move it to some common utils type package. I'm guessing this could be reused in the future.

okta/services/idaas/data_source_okta_identity_source_group_memberships.go
```diff
@@ -0,0 +1,215 @@
...
+		case ',':
+			if depth == 0 {
+				parts = append(parts, s[start:i])
+				start = i + 1
+			}
+		}
+	}
+	parts = append(parts, s[start:])
+	return parts
+}
```
dhiwakar-okta, line 199: If the above implementation changes and this method is unused, please consider removing it.

docs/data-sources/identity_source_groups.md
```diff
@@ -0,0 +1,37 @@
...
+- `identity_source_id` (String) ID of the identity source.
+- `external_id` (String) The external ID of the group in the identity source.
+
+### Optional
+
+- `id` (String) Okta-assigned ID of the group. May be provided to pre-populate or verify the Okta ID.
+
+### Read-Only
+
+- `id` (String) Okta-assigned ID of the group (populated after read).
```
dhiwakar-okta, line 37: What about [profile](https://developer.okta.com/docs/api/openapi/okta-management/management/tags/identitysource/other/getidentitysourcegroup#other/getidentitysourcegroup/t=response&c=200&path=profile) ?

okta/services/idaas/data_source_okta_identity_source_groups.go
```diff
@@ -0,0 +1,121 @@
...
+			"Unexpected DataSource Configure Type",
+			"Expected *config.Config, got something else. Please report this issue to the provider developers.",
+		)
+		return
+	}
+	d.Config = cfg
+}
+
+func (d *identitySourceGroupsDataSource) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	resp.Schema = schema.Schema{
```
dhiwakar-okta, line 60: What about [profile](https://developer.okta.com/docs/api/openapi/okta-management/management/tags/identitysource/other/getidentitysourcegroup#other/getidentitysourcegroup/t=response&c=200&path=profile) ?

docs/data-sources/identity_source_groups.md
```diff
@@ -0,0 +1,37 @@
...
+## Schema
+
+### Required
+
+- `identity_source_id` (String) ID of the identity source.
+- `external_id` (String) The external ID of the group in the identity source.
+
+### Optional
+
+- `id` (String) Okta-assigned ID of the group. May be provided to pre-populate or verify the Okta ID.
```
dhiwakar-okta, line 33: This is a computed value, a r[ead only value](https://developer.okta.com/docs/api/openapi/okta-management/management/tags/identitysource/other/getidentitysourcegroup#other/getidentitysourcegroup/t=response&c=200&path=id) right ? If yes, who is supposed to provide it ?

okta/services/idaas/data_source_okta_identity_source_sessions_generated.go
```diff
@@ -0,0 +1,173 @@
...
+		resp.Diagnostics.AddError(
+			"Unexpected DataSource Configure Type",
+			"Expected *config.Config, got something else. Please report this issue to the provider developers.",
+		)
+		return
+	}
+	d.Config = cfg
+}
+
+func (d *identitySourceSessionsDataSource) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
```
dhiwakar-okta, line 63: Is this data source for https://developer.okta.com/docs/api/openapi/okta-management/management/tags/identitysource/other/getidentitysourcesession ? 
If yes, the API requires 2 fields, namely `identitySourceId` & `sessionId`
It seems like you are just listing all sessions for the identity source & returning the most recent one. Why not simply mandate the session ID as well ?

okta/services/idaas/data_source_okta_identity_source_sessions_generated.go
```diff
@@ -0,0 +1,173 @@
...
+			"created": schema.StringAttribute{
+				MarkdownDescription: "The timestamp when the identity source session was created",
+				Computed:            true,
+			},
+			"import_type": schema.StringAttribute{
+				MarkdownDescription: "The type of import.  All imports are `INCREMENTAL` imports.",
+				Computed:            true,
+			},
+			"last_updated": schema.StringAttribute{
+				MarkdownDescription: "The timestamp when the identity source session was created",
```
dhiwakar-okta, line 85: You mean when the session was last updated ? 
This seems like a copy & paste mistake, please check other places where the description is incorrect :) .

okta/services/idaas/data_source_okta_identity_source_users_generated.go
```diff
@@ -0,0 +1,164 @@
...
+
+func (d *identitySourceUsersDataSource) Configure(_ context.Context, req datasource.ConfigureRequest, resp *datasource.ConfigureResponse) {
+	if req.ProviderData == nil {
+		return
+	}
+	cfg, ok := req.ProviderData.(*config.Config)
+	if !ok {
+		resp.Diagnostics.AddError(
+			"Unexpected DataSource Configure Type",
+			"Expected *config.Config, got something else. Please report this issue to the provider developers.",
```
dhiwakar-okta, line 68: Let's keep the error message the same across all resources, 
Check `okta/resources/safe_resource.go` and use the same error message, minus the stack trace since that won't be applicable here.

okta/services/idaas/resource_okta_identity_source_bulk_delete.go
```diff
@@ -0,0 +1,195 @@
...
+			"Expected *config.Config, got something else. Please report this issue to the provider developers.",
+		)
+		return
+	}
+	r.Config = cfg
+}
+
+func (r *identitySourceBulkDeleteResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Manages an Okta IdentitySourceBulkDelete resource.",
```
dhiwakar-okta, line 72: Does this description look right to you @aditya-okta ?

okta/services/idaas/resource_okta_identity_source_bulk_delete.go
```diff
@@ -0,0 +1,195 @@
...
+				},
+			},
+		},
+	}
+}
+
+func (r *identitySourceBulkDeleteResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	// Import ID format: {identity_source_id}/{session_id}/{id}
+	parts := strings.SplitN(req.ID, "/", 3)
+	if len(parts) != 3 || parts[0] == "" || parts[1] == "" || parts[2] == "" {
```
dhiwakar-okta, line 119: Ideally we should check if we have exactly 3 parts i.e 
len(parts) == 3 && parts[0] != "" && parts[1] != "" && parts[2] != ""

The current implementation would work if the ID is `abc/def/ghi/jkl/ghi`
The value is more than 3 parts. Yes it will eventually fail but the logic implemented here is a bit buggy.

okta/services/idaas/resource_okta_identity_source_bulk_delete.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	// Create returns no body (204); set ID from the plan field.
+	plan.ID = plan.SessionId
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
+}
+func (r *identitySourceBulkDeleteResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
+	resp.Diagnostics.AddWarning(
+		"Update Not Supported",
+		"This resource does not support in-place updates. Changes will require resource replacement.",
```
dhiwakar-okta, line 183: If the resource doesn't support in place updates and as you suggest requires replacement, can't you just re-create the resource in case of Update() ? 
Does that sound feasible @aditya-okta , @pranav-okta  ?

okta/services/idaas/resource_okta_identity_source_bulk_delete.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func (r *identitySourceBulkDeleteResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
+	var plan identitySourceBulkDeleteModel
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	identity_source_id := plan.IdentitySourceId.ValueString()
+	session_id := plan.SessionId.ValueString()
```
dhiwakar-okta, line 150: Use camelCase

okta/services/idaas/resource_okta_identity_source_bulk_delete.go
```diff
@@ -0,0 +1,195 @@
+// Code generated by tf-generator. DO NOT EDIT.
```
dhiwakar-okta, line 1: Why even have a resource when we can't read, delete or update it ? 
Also the above statement might not be technically correct.

Aren't the following endpoints managing the same resource ? 
Can't you get rid of the bulk upsert resource ? 

[https://developer.okta.com/docs/api/openapi/okta-management/management/tags/identitysource/other/uploadidentitysourcedataforupsert](https://developer.okta.com/docs/api/openapi/okta-management/management/tags/identitysource/other/uploadidentitysourcedatafordelete)

[https://developer.okta.com/docs/api/openapi/okta-management/management/tags/identitysource/other/uploadidentitysourcedataforupsert](https://developer.okta.com/docs/api/openapi/okta-management/management/tags/identitysource/other/uploadidentitysourcedataforupsert)

PR #2803, opened 2026-04-28
ci(release): add setup-terraform step to fix expired hc-install GPG key

aditya-okta:
## Summary
- The `hc-install` package embeds a HashiCorp GPG key to verify Terraform CLI downloads; that key has expired on Ubuntu runners where Terraform is not pre-installed
- Adding `hashicorp/setup-terraform` before the VCR smoke tests puts Terraform in PATH, so `hc-install` skips the download entirely — same pattern already used in `ci.yml`

## Test plan
- [ ] Re-run the release workflow after merging — VCR smoke tests should pass without the GPG key error

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Setup Terraform
         uses: hashicorp/setup-terraform@v4
         with:
-          terraform_version: "1.13.4"
+          terraform_version: "1.14.0"
 
       - name: Run VCR acceptance tests
         run: time make smoke-test-play-vcr-acc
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -30,6 +30,11 @@ jobs:
           go-version-file: "go.mod"
           cache: true
 
+      - name: Setup Terraform
+        uses: hashicorp/setup-terraform@v4
+        with:
+          terraform_version: "1.13.4"
```
pranav-okta: Why not the latest version v1.14.x?

dhiwakar-okta (approved):
LGTM 🚀

pranav-okta (approved):

Issue #2775, opened 2026-04-13
No users imported with `okta_group_memberships`


MScottBlake:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
        Before submitting a bug report, we ask that you first [search existing issues](https://github.com/okta/terraform-provider-okta/issues) and [pull requests](https://github.com/okta/terraform-provider-okta/pulls) to see if someone else may have experienced the same issue or may have already submitted a fix for it. This helps to keep all relevant information in one place, including any potential workarounds.

### A Note on Terraform Core Issues

We also ask that you consider whether your issue may be related to Terraform Core. If you are running into one of the following scenarios, we recommend [opening an issue](https://github.com/hashicorp/terraform/issues/new/choose) in the Terraform Core repository instead:

* [Configuration Language](https://developer.hashicorp.com/terraform/language) or resource ordering issues
* [State](https://developer.hashicorp.com/terraform/language/state) and [State Backend](https://developer.hashicorp.com/terraform/language/backend) issues
* [Provisioner](https://developer.hashicorp.com/terraform/language/resources/provisioners/syntax) issues
* [Registry](https://registry.terraform.io/) issues
* Issues that span resources across multiple providers

### Terraform Version & Okta Provider Version(s)

Terraform v1.14.8
on darwin_arm64
+ provider registry.terraform.io/okta/okta v6.9.0

### Affected Resource(s)

`okta_group_memberships`

### Can this be done in the Admin UI?

No

### Can this be done in the actual API call?

Yes

### Customer Information

Organization Name: Flywire
Paid Customer: yes

### Terraform Configuration

```tf
locals {
  g_id = "00gg1234567890123"
}

import {
  to = okta_group_memberships.test
  id = local.g_id
}

resource "okta_group_memberships" "test" {
  group_id        = local.g_id
  track_all_users = true
  users = [
    "00u12345678testuser1",
    "00u12345678testuser2",
    "00u12345678testuser3",
  ]
}
```

### Debug Output

[Terraform Plan](https://gist.github.com/MScottBlake/7d79c0e5e3be98fd6561f311be56ef7d)

[...]

PR #2810, opened 2026-05-08
GH-2775 | Import "users" When Importing okta_group_memberships

dhiwakar-okta:
`users` isn’t being imported when importing the resource `okta_group_memberships`.
This is because the import function currently just sets the group ID and not the users, which requires another API call.

Additionally this PR also fixes the VCR test `TestAccResourceOktaPushGroup_crud` by only setting app_config in the state file if at least one of the underlying fields has a non null value in the response.


pranav-okta (changes requested):
Could you add tests for the changes?

pranav-okta (approved):

Issue #2140, opened 2024-11-19
okta_auth_server_policy priority related repeat deployments


richard-collette-precisely:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

Terraform 1.9.8

### Affected Resource(s)

- okta_auth_server_policy

### Terraform Configuration Files

I'm using terraform CDK but the issue is with the underlying provider.

```typescript
    this.authServerPolicy = new AuthServerPolicy(
      this,
      `${configuration.label}_AccessPolicy`,
      {
        authServerId:
          this._options.authorizationServerConstruct.authorizationServer.id,
        clientWhitelist: [this._options.appOauth.id],
        description: configuration.label + " Access Policy",
        name: configuration.accessPolicyName,
        priority: 100,
        dependsOn: this._options.dependsOn
          ? [this._options.dependsOn]
          : undefined,
      },
    );
  }
```

### Debug Output

This is not debug output but sufficient to show the issue.  In these logs, you can see that an update is being performed to put the priority back to 100.

```bash
                      # okta_auth_server_policy.dis-dis-suiteadmin-accessmanagementapi-dev_ApplicationAccessPolicy_DIS-SuiteAdmin-AccessManagementApi-dev_AccessPolicy_5E343DCB (dis-dis-suiteadmin-accessmanagementapi-dev_ApplicationAccessPolicy/DIS-SuiteAdmin-AccessManagementApi-dev_AccessPolicy) will be updated in-place
                      ~ resource "okta_auth_server_policy" "dis-dis-suiteadmin-accessmanagementapi-dev_ApplicationAccessPolicy_DIS-SuiteAdmin-AccessManagementApi-dev_AccessPolicy_5E343DCB" {
                            id               = "00p28nks0t5Y2eA3G0h8"
                            name             = "SuiteAdmin-AccessManagementApi-dev"
                          ~ priority         = 1 -> 100
                            # (4 unchanged attributes hidden)
                        }
[...]

PR #2833, opened 2026-05-22
feat(auth_server_policy_priority): Add resource to manage policy priority ordering

### LINKED ISSUE #1198: bug(s) with resources having a `priority` setting
## Synopsis

We are looking closer at two bugs with resources that have a `priority` attribute. The bugs may be discrete or related. The resources that could possibility be affected are listed below.

## Bug 1 - Parallelism

The terraform runtime operates with a parallelism mechanism wherein it can make many API calls in parallel as it runs commands such as `plan` and `apply`. The [default is 10](https://www.terraform.io/cli/commands/apply#parallelism-n) and has a CLI flag of [-parallelism=n](https://www.terraform.io/cli/commands/apply#parallelism-n) to override the default. When unrestrained calls are made in this manner the API can 500 or 400 errors.

There is some feedback internally at Okta, that API resources having a `priority` property should not be called concurrently. Therefore the Okta Terraform provider needs to identify which resources can be affected by this. Then develop a mechanism to prevent the terraform runtime interacting with these resources in a parallel manner (beyond running the CLI in [-parallelism=n](https://www.terraform.io/cli/commands/apply#parallelism-n) mode). This might be tough to solve as the provider will be attempting to bend the TF runtime outside of its base behavior.

## Bug 2 - Ordering

A `priority` attribute implies ordering. However, it seems that if the all of the elements that are ordered are not present in an API call to Okta then the API result can be inconsistent. This is likely an input validation issue in the Okta Terraform Provider, and if this is the case, it could be handled with some additional API calls under the hood.

An example: imagine an [`okta_auth_server_policy_rule`](https://registry.terraform.io/providers/okta/okta/latest/docs/resources/app_signon_policy_rule) resource in a TF config (as reported in issue #713). Let's say that there is one rule being managed by the TF config but the policy has other rules not managed by TF.

| rule | priority | managed by TF |
|-----|--------|-----------------|
| A     | 1          | no                       |
| B     | 2         | yes                      |
| C     | 3         | no                       |

```terraform
resource "okta_auth_server_policy_rule" "test" {
  auth_server_id       = okta_auth_server.test.id
  policy_id            = okta_auth_server_policy.test.id
  status               = "ACTIVE"
  name                 = "B"
  priority             = 2
}
```
It appears that the initial `apply` when the `B` rule is created (and/or subsequent updates) can cause an Okta API error where the response is referencing the `priority` property of the TF resource. It seems as if the Okta Terraform provider first needs to inspect the parent of a resource (like `okta_auth_server_policy` is to `okta_auth_server_policy_rule`) and then gather all of the existing children (even those not manged by TF). Then any creates or updates need to be made in the context of all of the children associated with the parent. This is a hunch and needs to be verified and addressed.

## Resources
- okta_auth_server_policy
- okta_policy_rule_password
- okta_auth_server_policy_rule
- okta_policy_rule_signon
- okta_policy_rule_idp_discovery
- okta_user_schema_property
- okta_policy_password
- okta_user_schema
- okta_policy_signon
- okta_app_group_assignment
- okta_policy_rule_mfa
- okta_app_group_assignments
- okta_policy_mfa_default
- okta_app_signon_policy_rule
- okta_policy_password_default
- okta_policy_mfa
- okta_group_schema_property
- okta_auth_server_policy

gustavo-chan:
## Summary

Implements `okta_auth_server_policy_priority` resource to fix race condition in parallel terraform apply of `okta_auth_server_policy` resources (fixes #1198, #2140).

When multiple policies are created/updated in parallel, Okta normalizes priorities to contiguous integers in non-deterministic order. This coordinator resource serializes priority application (bottom-up, highest index first) to guarantee consistent ordering.

## Changes

- **New resource**: `okta_auth_server_policy_priority` — manages priority ordering only, does not create/delete policies
- Implements custom `ImportState` that lists all policies from API, sorts by priority, populates full state
- Read function detects drift: silently drops deleted policies, surfaces shortened list as plan diff
- Priority application bottom-up to avoid intermediate collisions
- Comprehensive docs with drift detection and lifecycle guidance
- Acceptance test with CRUD + import validation

## Testing

Acceptance tests pass against real Okta org
✓ Step 1: Create with order [first, second, third]
✓ Step 2: Update with reorder [third, first, second]
✓ Step 3: Import with custom ImportState

**Test pattern:** Creates three policies and okta_auth_server_policy_priority resource. Verifies priorities are correctly ordered using TestCheckResourceAttrPair. Import test ignores priorities field since API returns all policies on auth server, which may differ from test config.

## Notes

- Uses `terraform-plugin-framework` (not sdk/v2), matching pattern of `okta_app_signon_policy_rules`
- Individual `okta_auth_server_policy` resources should include `lifecycle { ignore_changes = [priority] }` to prevent conflicts
- Resource registered in `FWProviderResources()` per framework conventions


gustavo-chan:
Can I get a review and approval for this PR?

gustavo-chan:
Hi @dhiwakar-okta, can I get a review and approval for this PR?

okta/services/idaas/resource_okta_auth_server_policy_priority.go
```diff
@@ -0,0 +1,269 @@
...
+		policyID := policyIDs[i]
+		priority := int64(i + 1)
+
+		policy, _, err := client.AuthorizationServer.GetAuthorizationServerPolicy(ctx, authServerID, policyID)
+		if err != nil {
+			diags.AddError("Error reading authorization server policy",
+				fmt.Sprintf("Could not read policy '%s': %s", policyID, err))
+			return diags
+		}
+
```
nimeshsubramanianv2, line 230:
```suggestion
        // Skip update if priority is already correct
        if policy.PriorityPtr != nil && *policy.PriorityPtr == priority {
                continue
        }
```

No-op optimization missing — applyPriorities GETs + PUTs every policy even if the priority already matches. Consider skipping the update when *policy.PriorityPtr == priority:

nimeshsubramanianv2 (approved):

okta/services/idaas/resource_okta_auth_server_policy_priority.go
gustavo-chan: completed

gustavo-chan:
@aditya-okta, can you please review my PR? I would love to have this merged if possible.

PR #2839, opened 2026-05-26
fix: ignore window/arm from goreleaser

pranav-okta:
`Go 1.26.2` dropped support for the `windows/arm`, but   `.goreleaser.yml` was still trying to build for it, causing the error: go: unsupported   GOOS/GOARCH pair windows/arm.

Please see https://go.dev/doc/go1.26#windows


dhiwakar-okta (approved):
LGTM 🚀

PR #2843, opened 2026-05-27
fix: prevent nil pointer dereference in idp_saml, idp_social, and idp_oidc when accountLink.filter.groups is null

aditya-okta:
Fixes OKTA-1131393

## Problem

`okta_idp_saml`, `okta_idp_social`, and `okta_idp_oidc` resources crashed with a nil pointer dereference during `import`, `plan`, or `apply` on certain Okta cells.

The crash occurred in the Read function of all three resources:

```go
if idp.Policy.AccountLink.Filter != nil {
    setMap["account_link_group_include"] = utils.ConvertStringSliceToSet(
        idp.Policy.AccountLink.Filter.Groups.Include, // panics when Groups is nil
    )
}
```

The guard only checked `Filter != nil`, but not `Filter.Groups != nil`. On the affected cells, the API returns `accountLink.filter.groups = null` while `filter` itself is non-null — for example when the "Exclude admins" checkbox is unchecked:

```
"accountLink": {
  "filter": {
    "groups": null,
    "users": { "excludeAdmins": true }
  },
  "action": "AUTO"
}
```

## Fix

Added `&& idp.Policy.AccountLink.Filter.Groups != nil` to the nil guard in all three affected Read functions:

- `okta/services/idaas/resource_okta_idp_saml.go`
- `okta/services/idaas/resource_okta_idp_social.go`
- `okta/services/idaas/resource_okta_idp_oidc.go`

## Test

Added `TestAccResourceOktaIdpSaml_account_link_auto` — creates an `okta_idp_saml` with `account_link_action = "AUTO"` and no `account_link_group_include` (the customer's exact configuration), then imports it to exercise the Read path where the nil pointer was.

### DIFF
--- a/examples/resources/okta_idp_saml/account_link_auto.tf
+++ b/examples/resources/okta_idp_saml/account_link_auto.tf
@@ -1,0 +1,35 @@
+resource "okta_app_saml" "test" {
+  label                    = "testAcc_replace_with_uuid"
+  sso_url                  = "http://google.com"
+  recipient                = "http://here.com"
+  destination              = "http://its-about-the-journey.com"
+  audience                 = "http://audience.com"
+  subject_name_id_template = "$${user.userName}"
+  subject_name_id_format   = "urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress"
+  response_signed          = true
+  signature_algorithm      = "RSA_SHA256"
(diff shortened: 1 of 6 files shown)


dhiwakar-okta (approved):
LGTM 🚀

PR #2871, opened 2026-06-26
feat: fix papercuts around domain verifications

### LINKED ISSUE #1909: okta_email_domain_verification throws 400 error if email domain is already verified
### Community Note

### Terraform Version
1.5.7

### Affected Resource(s)

- okta_email_domain_verification

```hcl
resource "okta_email_domain_verification" "vanity_email_address_verification" {
  email_domain_id = okta_email_domain.vanity_email_address.id
}
```

### Debug Output

From Terraform: `Error: failed to verify email domain: 400 Bad Request`

### Expected Behavior

Resource should check to see if the email domain is already verified and if so return as successful (Maybe do aGET call to check the status of the email domain first before trying to verify)

Yes

Yes

### Actual Behavior

If the email Domain is already verified (via manually in the UI or an API call) the resource will throw a 400 with the following error message: `"errorSummary": "status: Cannot validate email domain in current status."`

Since this resource currently does not have the ability to be imported we will always get this error if the email has already been verified.

ohookins:
There are two problems with the `okta_domain_verification` and `okta_email_domain_verification` resources right now:

1. If the custom domain / email domain is already verified by DNS, attempting to create these resources (by triggering another verification) will fail with a 400 error. 
2. There's no way to import the current verification status of the domain.

I've attempted to solve both of these problems by
1. Querying the domain verification status before attempting to verify again, so avoiding triggering an API request which unfortunately is not idempotent.
2. Allowing the user to import the domain verification and avoid case (1) altogether if desired.

Fixes #1909

### DIFF
--- a/docs/resources/domain_verification.md
+++ b/docs/resources/domain_verification.md
@@ -16,7 +16,7 @@ resource "okta_domain" "example" {
 }
 
 resource "okta_domain_verification" "example" {
-  domain_id = okta_domain.test.id
+  domain_id = okta_domain.example.id
 }
 ```
 
@@ -30,5 +30,13 @@ resource "okta_domain_verification" "example" {
 ### Read-Only
 
 - `id` (String) The ID of this resource.
+
+## Import
+
+Import is supported using the following syntax:
+
+```shell
+terraform import okta_domain_verification.example <domain_id>
+```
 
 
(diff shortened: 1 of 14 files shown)


ohookins:
@pranav-okta is anybody able to review this?

Issue #2356, opened 2025-06-18
Custom Roles using `okta.workflows.*` have permanent drift due to undocumented permission changes


delize:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
        Before submitting a bug report, we ask that you first [search existing issues](https://github.com/okta/terraform-provider-okta/issues) and [pull requests](https://github.com/okta/terraform-provider-okta/pulls) to see if someone else may have experienced the same issue or may have already submitted a fix for it. This helps to keep all relevant information in one place, including any potential workarounds.

### A Note on Terraform Core Issues

We also ask that you consider whether your issue may be related to Terraform Core. If you are running into one of the following scenarios, we recommend [opening an issue](https://github.com/hashicorp/terraform/issues/new/choose) in the Terraform Core repository instead:

* [Configuration Language](https://developer.hashicorp.com/terraform/language) or resource ordering issues
* [State](https://developer.hashicorp.com/terraform/language/state) and [State Backend](https://developer.hashicorp.com/terraform/language/backend) issues
* [Provisioner](https://developer.hashicorp.com/terraform/language/resources/provisioners/syntax) issues
* [Registry](https://registry.terraform.io/) issues
* Issues that span resources across multiple providers

### Terraform Version & Okta Provider Version(s)

Terraform v1.11.3
on Hashicorp Cloud / Linux
+ provider registry.terraform.io/providers/okta/okta/ v4.19.0

### Affected Resource(s)

https://registry.terraform.io/providers/okta/okta/latest/docs/resources/admin_role_custom

`okta_admin_role_custom`

### Can this be done in the Admin UI?

Yes

### Can this be done in the actual API call?

Yes

### Customer Information

Organization Name: eqtpartners
Paid Customer: yes

### Terraform Configuration

```tf
module "tech_support_workflow_resource_set" {
  source  = "app.terraform.io/organization_id/okta-resource-set-module/eqt"
  version = "0.5.29"
  all_workflows = true
  resource_set_enabled     = true
  resource_set_label       = "Tech Support Workflow Resource Set"
[...]

PR #2883, opened 2026-07-07
GH-2356 | Custom Roles using okta.workflows.* Has Permanent Drift Due to Undocumented Permission Changes

dhiwakar-okta:
Migrating to V6 SDK to address the issue reported in https://github.com/okta/terraform-provider-okta/issues/2356


dhiwakar-okta:
Not required.

Issue #1218, opened 2022-07-25
okta_user_group_memberships - track_all_groups


exitcode0:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

tl;dr - We should add `track_all_groups` to `okta_user_group_memberships` similar to `okta_group_memberships`

Currently the `okta_user_group_memberships` only manages user memberships it created, this allows it to co-exist with other management strategies such as click-ops or Okta Workflows

However, there are use-cases for wanting to `okta_user_group_memberships` to be the Definitive source of truth for all of a user's Okta group memberships, e.g service accounts.

Having `track_all_users` `okta_group_memberships` would allow me to prevent group membership drift on service account group memberships and allow me to be confident that my service accounts are still least privileged
`track_all_users` would also allows for some interesting Terraform modules, For example, create a Terraform provider that accepts a group ID & justification string, you can then use a variable validator to ensure the justification string matches the expected format (e.g contains a ticketing system URL)

The only current workaround that I know of is to manage all groups with Terraform and use `track_all_users` on every `okta_group_memberships` but this is unrealistic as in my environment some groups are managed by systems outside of Terraform

### New or Affected Resource(s)

- `okta_user_group_memberships`

### Potential Terraform Configuration

```hcl
resource "okta_user" "this" {
  login      = "example@acme.com"
  email      = "example@acme.com"
  first_name = "example"
  last_name  = "service-account"
  status     = "ACTIVE"
}

resource "okta_user_group_memberships" "this" {
  user_id = okta_user.this.id
  track_all_groups = true
  groups = [
    okta_group.service_accounts.id,
    okta_group.google_users.id,
  ]
}
```

### References

- #0000

monde:
Thanks @exitcode0 , this sounds good to me.

monde:
Okta internal reference: https://oktainc.atlassian.net/browse/OKTA-524620

exitcode0:
@monde 
Just wanting to clarify if this was closed by stale-bot or if this is considered a won't fix?

monde:
Keeping open.

exitcode0:
[...]

PR #2888, opened 2026-07-11
GH-1218 | Migrate to V6 SDK and Add track_all_groups to okta_user_group_memberships

dhiwakar-okta:
- Migrates `okta_user_group_memberships` and `okta_group_memberships` (and their shared helpers in `group.go`) from the V2 SDK (`sdk.Client`) to the V6 SDK (`v6okta.APIClient`), aligning with the provider's ongoing SDK migration.

- Implements the `track_all_groups` feature requested in #1218 for `okta_user_group_memberships`.

- Fixes a latent bug where removing a managed group externally caused a full resource recreate instead of a targeted update.


okta/services/idaas/resource_okta_user_group_memberships.go
```diff
@@ -19,7 +19,16 @@ func resourceUserGroupMemberships() *schema.Resource {
...
+This resource allows you to bulk manage groups for a single user, independent of
+the user schema itself. This allows you to manage group membership in terraform
+without overriding other automatic membership operations performed by group rules
+and other non-managed actions.
+**Important**: The default behavior of the resource is to only maintain the
+state of group ids that are assigned to it. This behavior will signal drift only if
+those groups stop being part of the user's memberships. If the desired behavior is
+to track all groups that are added/removed from the user make use of the
+'track_all_groups' argument with this resource.`,
 		Schema: map[string]*schema.Schema{
```
exitcode0, line 32: It looks like you missed adding the attribute to the resource's schema, though `resourceUserGroupMembershipsRead()` reads it. `d.Get` on an undeclared key returns nil, so the `.(bool)` assertion panics on every read.

```suggestion
		Schema: map[string]*schema.Schema{
			"track_all_groups": {
				Type:        schema.TypeBool,
				Optional:    true,
				Default:     false,
				Description: "The resource concerns itself with all groups added/deleted to the user; even those managed outside of the resource.",
			},
```

PR #2906, opened 2026-07-30
fix: normalize chains JSON keys during plan phase (OKTA-1184047)

aditya-okta:
**Problem**:  When a user configures chains with jsonencode({...}) where HCL attribute order puts userVerification before method, Terraform preserves that source ordering in the JSON string. After apply, the
  provider's Read path calls NormalizeDataJSON which re-marshals the chain through a Go map — sorting all keys alphabetically. The Framework then compares the planned value (non-alphabetical key order)
  against the post-apply state value (alphabetical key order) and raises:

  Error: Provider produced inconsistent result after apply
  .rule[1].chains[0]: was cty.StringVal("{...\"userVerification\":\"OPTIONAL\",\"method\":\"push\"...}"),
                     but now cty.StringVal("{...\"method\":\"push\",\"userVerification\":\"OPTIONAL\"...}")

**Soln**: Add ChainsPlanModifier, a planmodifier.List on the chains attribute that normalizes JSON key order during the plan phase, before apply. This makes the planned value match the post-apply state value
  that NormalizeDataJSON on Read already produces.

  The modifier validates each chain string with json.Unmarshal and surfaces an AddAttributeError diagnostic for invalid JSON rather than silently corrupting the value to "{}".

**Changes**
  - resource_okta_app_signon_policy_rules.go — new ChainsPlanModifier wired into the chains schema attribute's PlanModifiers
  - resource_okta_app_signon_policy_rules_test.go — unit test (TestChainsPlanModifier) covering sorted keys, unsorted keys, nested structures, and invalid JSON; acceptance test
  (TestAccResourceOktaAppSignOnPolicyRules_chains_misaligned_keys) with a PlanOnly idempotency step that would have failed before the fix
  - examples/resources/okta_app_signon_policy_rules/chains_misaligned_keys.tf — example config reproducing the customer scenario
  - test/fixtures/vcr/idaas/.../oie-00.yaml — VCR cassette for the acceptance test

### DIFF
--- a/okta/services/idaas/resource_okta_app_signon_policy_rules.go
+++ b/okta/services/idaas/resource_okta_app_signon_policy_rules.go
@@ -159,6 +159,48 @@ func (m reauthFrequencyModifier) PlanModifyString(ctx context.Context, req planm
 	}
 }
 
+// ChainsPlanModifier normalizes the JSON key order of each chains element during
+// planning so the plan value always matches the post-apply canonical form.
+type ChainsPlanModifier struct{}
+
+func (m ChainsPlanModifier) Description(_ context.Context) string {
+	return "Normalizes JSON key order in chains elements to prevent inconsistent result after apply"
+}
+
+func (m ChainsPlanModifier) MarkdownDescription(ctx context.Context) string {
+	return m.Description(ctx)
+}
+
+func (m ChainsPlanModifier) PlanModifyList(ctx context.Context, req planmodifier.ListRequest, resp *planmodifier.ListResponse) {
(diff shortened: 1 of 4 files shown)


okta/services/idaas/resource_okta_app_signon_policy_rules.go
dhiwakar-okta, line 824: `constraints` is the same "list of JSON-encoded strings" shape and is also re-serialized on Read. Maybe not now but this might need to be addressed in the future if someone reports a similar issue with this field.

dhiwakar-okta (approved):
Approved with comments.

PR #2915, opened 2026-08-05
Revert "feat: update okta-trusted-origin resource (#2884)"

pranav-okta:
This reverts commit 5a1e9ed6109b6fce657375613337865e213d6581.

PR #2884, opened 2026-07-08
feat: update okta-trusted-origin resource

pranav-okta:
JIRA: https://oktainc.atlassian.net/browse/OKTA-1218464
- Migrates okta_trusted_origin from SDKv2 to Terraform Plugin Framework using the generated implementation. Adds a generated data source (**data_source_okta_trusted_origin_generated.go**), replaces the SDKv2 resource with the Framework version, and updates VCR cassettes for the existing CRUD acceptance test.

PR #2932, opened 2026-08-18
Reapply "feat: update okta-trusted-origin resource (#2884)" (#2915)

pranav-okta:
This reverts commit da93ac0ac0d565d8188bbb65b11c8a01959f15ca.


docs/resources/trusted_origin.md
dhiwakar-okta, line 1: This markdown file needs to include an example illustrating how to migrate from v6.X.X to v7.X.X in case there is any breaking change in the resource, which I believe there is.
It should show both the old definition and new definition, something like 

_In v7.x.x, okta_trusted_origin needs to be defined differently than in v6.x.x, as illustrated by the following example:_

_v6.x.x_ 
```
resource "okta_trusted_origin" "example" {
  name   = "example"
  origin = "https://example.com"
  scopes = ["CORS"]
}
```
 _v7.x.x_
```
resource "okta_trusted_origin" "example" {
  name   = "example"
  origin = "https://example.com"
  scopes {
    type = "CORS"
  }
}
```

Follow this template for other resources in other PRs having similar breaking changes.

dhiwakar-okta (dismissed):
Approved with comments.

docs/resources/trusted_origin.md
pranav-okta: This is done.

dhiwakar-okta (dismissed):
LGTM 🚀

dhiwakar-okta (approved):
LGTM 🚀

PR #2942, opened 2026-08-28
feat: allow updating default rule of signon policy

### LINKED ISSUE #2788: Cannot modify default rule for default global session policy
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* If an issue is assigned to a user, that user is claiming responsibility for the issue.
* [OKTA support]() Customers can ask to reach out to Okta Developer Support Engineer/Customer Success Engineer to expedite investigation and resolution of this issue.

### Description

Default global session policy and its default rules can be imported into Terraform to bring these resources under Terraform management.

However, its not supported by okta terraform provider to update the default global session policy's default rule like MFA requirement or session configuration.

```
okta_policy_rule_signon.default: Modifying... [id=****************]
╷
│ Error: failed to update sign-on policy rule: Default Rule is immutable
│
│   with okta_policy_rule_signon.default,
│   on main.tf line 1, in resource "okta_policy_rule_signon" "default":
│    1: resource "okta_policy_rule_signon" "default" {
│
╵
```

UI and API allow to update the default rule.

As per [API documentation](https://developer.okta.com/docs/api/openapi/okta-management/management/tags/policy/section/rule-priority#section/Default-rules), it should be possible to update the default rule, with few restrictions:
> Only the default policy contains a default rule. In Identity Engine, you can't delete default rules, but can edit them, except for:
The property usePersistentCookie of the default global session policy's default rule, which is read-only.

### New or Affected Resource(s)

* okta_policy_rule_signon

### Customer Information

Organization Name: <your-org-name>
Paid Customer: <yes/no>

### Potential Terraform Configuration

Import the default rule to manage as terraform resource.

`terraform import okta_policy_rule_signon.default <policy_id>/<rule_id>`

```
variable "default_signon_policy_id" {
  type        = string
  description = "Default SignOn Id"
}
```

```
resource "okta_policy_rule_signon" "default" {
  policy_id = var.default_signon_policy_id
  name      = "Default Rule"
  status    = "ACTIVE"

  mfa_required = true
  session_lifetime = 1440
}
```

### References

[Cannot modify default MFA policy](https://github.com/okta/terraform-provider-okta/issues/191)
[Okta MFA Policy Default Rules](https://github.com/okta/terraform-provider-okta/issues/744)

pranav-okta:
Fixes #2788

### DIFF
--- a/docs/resources/policy_rule_mfa.md
+++ b/docs/resources/policy_rule_mfa.md
@@ -152,6 +152,7 @@ resource "okta_policy_rule_mfa" "example" {
 ### Read-Only
 
 - `id` (String) The ID of this resource.
+- `system` (Boolean) Whether this is the system (default) rule of its policy. Okta creates and deletes default rules itself; Terraform can update one but can not create, rename or delete it.
 
 <a id="nestedblock--app_exclude"></a>
 ### Nested Schema for `app_exclude`
--- a/okta/services/idaas/resource_okta_policy_rule_mfa.go
+++ b/okta/services/idaas/resource_okta_policy_rule_mfa.go
@@ -19,6 +19,11 @@ func resourcePolicyMfaRule() *schema.Resource {
 		Importer:      createPolicyRuleImporter(),
 		Description:   "Creates an MFA Policy Rule. This resource allows you to create and configure an MFA Policy Rule.",
 		Schema: buildRuleSchema(map[string]*schema.Schema{
+			"system": {
+				Type:        schema.TypeBool,
+				Computed:    true,
+				Description: "Whether this is the system (default) rule of its policy. Okta creates and deletes default rules itself; Terraform can update one but can not create, rename or delete it.",
+			},
 			"enroll": {
 				Type:        schema.TypeString,
 				Default:     "CHALLENGE",
@@ -68,7 +73,7 @@ func resourcePolicyMfaRuleRead(ctx context.Context, d *schema.ResourceData, meta
 	if err != nil {
 		return diag.Errorf("failed to sync MFA policy rule: %v", err)
 	}
-	if (rule.Conditions.App) != nil {
+	if rule.Conditions != nil && rule.Conditions.App != nil {
 		if len(rule.Conditions.App.Include) != 0 {
 			_ = d.Set("app_include", flattenApps(rule.Conditions.App.Include))
 		}
@@ -92,7 +97,9 @@ func resourcePolicyMfaRuleUpdate(ctx context.Context, d *schema.ResourceData, me
 }
 
 func resourcePolicyMfaRuleDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-	err := deleteRule(ctx, d, meta, false)
+	// checkIsSystemPolicy is true so that a policy's default rule, which Okta will not
+	// let anyone delete, is dropped from state instead of failing the destroy.
+	err := deleteRule(ctx, d, meta, true)
 	if err != nil {
 		return diag.Errorf("failed to delete MFA policy rule: %v", err)
 	}
--- a/okta/services/idaas/resource_okta_policy_rule_sign_on.go
+++ b/okta/services/idaas/resource_okta_policy_rule_sign_on.go
@@ -19,8 +19,17 @@ func resourcePolicySignOnRule() *schema.Resource {
 		UpdateContext: resourcePolicySignOnRuleUpdate,
 		DeleteContext: resourcePolicySignOnRuleDelete,
 		Importer:      createPolicyRuleImporter(),
-		Description:   "Creates a Sign On Policy Rule. In case `Invalid condition type specified: riskScore.` error is thrown, set `risc_level` to an empty string, since this feature is not enabled.",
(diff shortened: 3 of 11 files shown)


okta/services/idaas/resource_okta_policy_rule_mfa.go
```diff
@@ -19,6 +19,11 @@ func resourcePolicyMfaRule() *schema.Resource {
 		Importer:      createPolicyRuleImporter(),
 		Description:   "Creates an MFA Policy Rule. This resource allows you to create and configure an MFA Policy Rule.",
 		Schema: buildRuleSchema(map[string]*schema.Schema{
+			"system": {
```
dhiwakar-okta, line 22: Did you try upgrading from v7.0.0 to the branch containing your changes ? 
Are we seeing any one time diffs for resources ?

okta/services/idaas/resource_okta_policy_rule_sign_on.go
```diff
@@ -19,8 +19,17 @@ func resourcePolicySignOnRule() *schema.Resource {
...
 		DeleteContext: resourcePolicySignOnRuleDelete,
 		Importer:      createPolicyRuleImporter(),
-		Description:   "Creates a Sign On Policy Rule. In case `Invalid condition type specified: riskScore.` error is thrown, set `risc_level` to an empty string, since this feature is not enabled.",
+		Description: `Creates a Sign On Policy Rule. In case 'Invalid condition type specified: riskScore.' error is thrown, set 'risc_level' to an empty string, since this feature is not enabled.
+A policy's default rule (the rule named 'Default Rule') can be imported and updated as a custom rule, but it can not be created, renamed or deleted.
+Okta manages these fields on a default rule, so values configured for them are ignored: 'priority', 'network_connection', 'network_includes',
+'network_excludes', 'users_excluded' and 'session_persistent'.
+Removing a default rule from your configuration drops it from Terraform state without deleting it from Okta.`,
 		Schema: buildRuleSchema(map[string]*schema.Schema{
+			"system": {
```
dhiwakar-okta, line 28: Did you try upgrading from v7.0.0 to the branch containing your changes ?
Are we seeing any one time diffs for resources ?

docs/resources/policy_rule_mfa.md
```diff
@@ -152,6 +152,7 @@ resource "okta_policy_rule_mfa" "example" {
 ### Read-Only
 
 - `id` (String) The ID of this resource.
+- `system` (Boolean) Whether this is the system (default) rule of its policy. Okta creates and deletes default rules itself; Terraform can update one but can not create, rename or delete it.
```
dhiwakar-okta, line 155: Is there any new VCR test that you're adding for testing this new field ? It seems like you've re-recorded existing tests which is fine but isn't there supposed to be a new test as well ?

dhiwakar-okta (approved):
Approved with comments.

sdk/v2_userCondition.go
```diff
@@ -2,7 +2,12 @@
 package sdk
 
 type UserCondition struct {
-	Exclude []string `json:"exclude,omitempty"`
```
dhiwakar-okta, line 5: Does removing `omitempty` affect other resources ? 
`okta_app_signon_policy_rule`, okta_auth_server_policy_rule` and `okta_app_signon_policy_rules` would also be impacted by this, right ?
They also reference this struct.
Also if it doesn't make sense of `Exclude` to have the omitempty tag, then it shouldn't be there of `Include` either, right ?

okta/services/idaas/resource_okta_policy_rule_sign_on.go
pranav-okta: I checked into this since `okta_policy_rule_signon/_mfa` aren't new resources, so plenty of people already have them in state. So, this won't cause any conflicts.

sdk/v2_userCondition.go
pranav-okta: - The Policies API from Okta always sends  `"users":{} → "users":{"exclude":[]}` empty exclude. Removing the `omitempty` will fix this issue because the TF import performs a byte-by-byte check for drift, and our SDK does not return the `exclude` param when there is no value.
- Left `Include` alone on purpose — went through all 677 recorded test fixtures in the repo, and Okta has never once sent back an empty include array, only empty exclude. So there's nothing there to actually fix.

docs/resources/policy_rule_mfa.md
pranav-okta: This is a `Computed` attribute that we can not create or update. We need this to check whether the policy/rule is a default during import. I think we don't need any extra tests for this.

There is another problem with writing the test. We don't actually hit the API but just record/replay the cassette. To import the default rule, the test needs to find its ID first, and the only code path that performs that lookup uses a client built outside the recording system — so it always tries to make a real network call, whether recording or not. Fixing this means fixing the test setup to find the ID through itself which is beyond the scope of the bug.

dhiwakar-okta (approved):
LGTM 🚀

PR #2952, opened 2026-09-14
fix: paginate user groups correctly in okta_user_group_memberships

### LINKED ISSUE #2951: okta_user_group_memberships fails for users with more than 400 groups: pagination loop overwrites the previous page
### Community Note

* Please vote on this issue by adding a 👍 reaction to the original issue to help the community and maintainers prioritize this request.

I have a fix ready and will open a pull request referencing this issue.

### Terraform Version & Okta Provider Version(s)

```
OpenTofu v1.11.6
on linux_amd64
+ provider registry.terraform.io/providers/okta/okta/ v6.12.0
```

The affected code is unchanged in v6.13.0, v6.14.0, v6.15.0, v7.0.0 and current `master`.

### Affected Resource(s)

* `okta_user_group_memberships`

```hcl
data "okta_user" "approver" {
  search {
    name  = "profile.email"
    value = "someone@example.com" # a user who is a member of more than 400 groups
  }
}

resource "okta_group" "access_approver" {
  name = "Application: Example - Access Approver"
}

resource "okta_user_group_memberships" "approver" {
  user_id = data.okta_user.approver.id
  groups  = [okta_group.access_approver.id]
}
```

### Expected Behavior

Apply succeeds. The user is a member of the group (Okta's `GET /api/v1/groups/{id}/users` and a fully paginated `GET /api/v1/users/{id}/groups` both list it), so the post-create membership check should pass. On refresh, the resource should stay in state.

### Actual Behavior

Create fails every time, deterministically:

```
Error: user (00u…) did not have expected group memberships after multiple checks

  with okta_user_group_memberships.approver,
  on main.tf line 101, in resource "okta_user_group_memberships" "approver":
```

The `PUT /api/v1/groups/{id}/users/{userId}` itself succeeds (the user is added, or was already a member). Because `resourceUserGroupMembershipsRead` runs the same check, an existing resource for such a user is also silently dropped from state on refresh (`d.SetId("")`), so every subsequent plan shows it as "will be created" and every apply fails again.

### Steps to Reproduce

1. Pick a user who is a member of more than 400 groups, so `GET /api/v1/users/{id}/groups` spans at least three pages of 200.
2. Add that user to a group whose id sorts into the first `len(page 3)` entries of page 2 of that listing.
3. `terraform apply` fails with the error above. Repeating the apply fails identically, so this is not eventual-consistency lag.

### Root cause

`checkIfUserHasGroups` in `okta/services/idaas/resource_okta_user_group_memberships.go` declares one `nextUserGroups []*sdk.Group` outside the pagination loop and passes `&nextUserGroups` to `resp.Next` for every page:

```go
var nextUserGroups []*sdk.Group

for resp.HasNextPage() {
    resp, err = resp.Next(ctx, &nextUserGroups)
    ...
    userGroups = append(userGroups, nextUserGroups...)
}
```

`encoding/json` resets the slice length to zero and then decodes into the existing backing array. Because the elements are pointers, the decoder writes into the `sdk.Group` structs that `userGroups` already references from the previous page. So decoding page 3 overwrites the first `len(page 3)` groups of page 2, and those groups are never seen by the check. Users with exactly two pages (201–400 groups) are unaffected, which is why this is easy to miss.

Verified against a real org: two users with 473 and 489 groups, target group at page 2 index 8 and 17 respectively, page 3 sizes 73 and 89. Both fail; walking the same endpoint with curl and following `Link: rel="next"` finds the group.

The sibling resource `okta_group_memberships` already avoids this by resetting the slice (`groupUsers = nil`) before each `resp.Next` call.

### Important Factoids

* Only affects users with more than 400 groups (three or more pages).
* #1371 added the pagination loop that contains this bug.

### References

* #1371

fixe:
Fixes #2951

## Summary

`okta_user_group_memberships` fails to create, and is silently dropped from state on refresh, for any user whose `GET /api/v1/users/{id}/groups` listing spans three or more pages (more than 400 groups):

```
Error: user (00u…) did not have expected group memberships after multiple checks
```

`checkIfUserHasGroups` declares one `nextUserGroups []*sdk.Group` outside the pagination loop and passes `&nextUserGroups` to `resp.Next` for every page. `encoding/json` decodes into the existing backing array, and because the elements are pointers it writes into the `sdk.Group` structs that `userGroups` already references from the previous page. Decoding page 3 therefore overwrites the first `len(page 3)` groups of page 2, and those groups are never seen by the check. Users with exactly two pages are unaffected, which is why this has gone unnoticed.

The fix resets the slice before each `resp.Next` call, the same pattern `okta_group_memberships` already uses (`groupUsers = nil`).

## Changes

- `okta/services/idaas/resource_okta_user_group_memberships.go`: reset `nextUserGroups` before decoding each page.
- `okta/services/idaas/resource_okta_user_group_memberships_internal_test.go`: new `package idaas` unit test. An `httptest` server serves three pages (200 / 200 / 73 groups) with Okta-style `Link: rel="next"` headers, and the test asserts that groups on every page are found and that missing groups are still reported as missing. Without the fix, the "start of second page" and "one group per page" cases fail.
- `CHANGELOG.md`: entry under Unreleased / BUG FIXES.

## Checklist

- [x] Test coverage that fails if the fix is removed. The existing tests in this package are VCR acceptance tests, and a user with 400+ groups is not practical to record into a cassette, so this uses a local mock server against the unexported check instead. Happy to restructure if you would prefer a different shape.
- [x] No documentation changes needed; behaviour is unchanged for users with up to 400 groups.
- [x] `gofmt`, `go vet`, `tfproviderlint -c 1` clean; package compiles under the `unit` tag used by the VCR CI job.

## Verification

Reproduced against a real org: two users with 473 and 489 groups, target group at page 2 index 8 and 17, page 3 sizes 73 and 89. Both fail identically on every apply, and walking the same endpoint with curl while following `Link: rel="next"` finds the group.


rasta-rocket:
@fixe just to let you know that there was already a PR opened about that waiting for review: https://github.com/okta/terraform-provider-okta/pull/2898

